PR #51, opened 2019-04-07
Add Philipp John /uses

JohnnyDevNull:


igauravsehrawat:
@thisisabdus I think you should ask for description as it adds more value to that uses page. I am doing it on other PR, after that I will merge. What say?

thisisabdus:
@igauravsehrawat Hey :wave: Yeah, I could've asked more description in the /uses page. But I think whatever tool developers are using is more important then writing about the tools description as one could easily google about the use-case of those tools :) 
It's their personal site. I can't ask them what to write and what not to :sweat_smile: But Yeah, If they write little description, some people might get benefited out of that.

JohnnyDevNull:
@igauravsehrawat can you give me some example what you mean with description in that case please?

Do you mean some kind of description when I more prever these or that tool like VSCode vs WebStorm?

PR #133, opened 2020-01-08
added self already, just x'ing off list

davidosomething:


davidosomething:
might be better to move this list to the wiki

PR #134, opened 2020-01-08
Add @nomangul

NomanGul:


src/data.js
```diff
@@ -447,6 +447,23 @@ const pages = [
       'Node',
       'Vue'
     ],
+  },
+  {
+    name: 'Noman Gul',
+    description:
+      'Jr Software Engineer, Tutorial Maker, Biryani Lover',
+    url: 'https://howivscode.com/NomanGul',
```
armujahid: 1. URL must ends with `/uses`
2. This list is not specific to vscode

NomanGul: @armujahid oh, ok... Thanks :slightly_smiling_face:

PR #138, opened 2020-01-08
Issue 120: Added filter buttons for computers and phones

byurhanbeyzat:
- Fixed phone to iphone from apple
- Added computer and phones to GraphQL API 
- Added queries and the buttons to the Topics bar


wesbos:
thank a ton!

PR #143, opened 2020-01-09
Remove pointer style on non-interactive tags

MichaelE1:
Just removing the cursor pointer styling on the tags for each person - that don't function as buttons. Cheers.


wesbos:
thanks!

PR #169, opened 2020-01-09
Add OP to uses

xi-osman:


wesbos:
please add a link to your /uses page and also use an emoji for the country

PR #176, opened 2020-01-09
Added: sort people in alphabetical order.

chuckrincon:


wesbos:
We intentionally randomize them on each build so we don't hurt any feelings, so I appreciate the PR but we don't want to sort them :)

PR #177, opened 2020-01-09
Update data.js

GantMan:


src/data.js
```diff
@@ -34,6 +34,32 @@ const pages = [
...
   },
+  {
+    name: 'Gant Laborde',
+    description: 'Speaker, GDE Web/ML, Podcaster, Trainer, Speaker, Author, Podcaster',
+    url: 'http://gantlaborde.com/uses/',
+    twitter: '@gantlaborde',
+    emoji: '🔥',
+    country: 'US',
+    computer: 'windows',
+    phone: 'ios',
```
tlhunter: I believe this breaks the page rendering and should instead be `iphone`.

PR #187, opened 2020-01-09
Added Carlos Andres Charris S

Gdaimon:


wesbos:
thanks -could you also add yourself to data.js for the new uses.tech website? https://github.com/wesbos/awesome-uses/blob/website/src/data.js

PR #222, opened 2020-01-09
Add Tom Hazledine

tomhazledine:
Such a great idea; got totally nerd-sniped by this one! Created a `\uses` page just for this. 👍


wesbos:
added

PR #242, opened 2020-01-10
Add David Llop info

Lloople:


src/data.js
```diff
@@ -731,6 +731,29 @@ const pages = [
       'Blogger',
     ],
   },
+  {
+    name: 'David Llop',
+    description: 'Full stack developer from Girona. Open Source contributor. Always Learning.',
+    url: 'https://davidllop.com',
```
Jamesking56: Not a `/uses` page?

Lloople: Whoops, forgot about that, thank you!

PR #243, opened 2020-01-10
Fix @gantlaborde phone and add @flexbox

flexbox:
**before**
image

**after**
image


wesbos:
thnaks already fixed this

Issue #156, opened 2020-01-09
Trim trailing / on /uses/


wesbos:
With the display URL, some have a trailing `/` which is no way to live your life.

If the url ends in `/`, can we trim it from the display? 

As the code is pretty minor, I'm going to leave this one for someone who wants to tip their toes into the whole github fork/PR workflow. 

If you are already experienced, Please leave this one for someone who wants a good first issue :)

yosefanajjar:
I'd be happy to contribute to a Webos repo, just did it in the simplest way I could, I'll make a PR now

PR #262, opened 2020-01-10
Trim Display URL

yosefanajjar:
Relates #156 

1. trim the trailing slash from the display URL
2. Added prop type for `person.country`


wesbos:
thanks!

PR #264, opened 2020-01-10
Make sure long links does not break the design

Dinoshauer:
Fixes long links breaking the design with ellipsis

image


wesbos:
thanks a bundle

PR #279, opened 2020-01-10
feat: add action to validate data.js

### LINKED ISSUE #141: Reject PR if invalid data
Need some way to reject the PR if the submitted data doesn't meet the shape and requirements of what we want.

Maybe convert to Typescript? 

Or can we do it with with PropTypes? 

Seems like a good use case for github actions

iamandrewluca:
Closes #141

### DIFF
--- a/.github/workflows/data-validate.yml
+++ b/.github/workflows/data-validate.yml
@@ -1,0 +1,25 @@
+name: Validate data.js
+
+on:
+  pull_request:
+    paths: src/data.js
+
+env:
+  CI: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


wesbos:
thanks - do you have example of this on your repo? Will it show up with the netlify stuff in a PR?

iamandrewluca:
Yes, it will be like Netlify stuff

wesbos:
thanks! testing it myself and it seems e._original should be e._object? I can fix it if so

iamandrewluca:
I also observed this, but it worked. I think that Joi static types do not reflect real data.

iamandrewluca:
As I see action is working well

wesbos:
awesome! Im working on that now - adding some better errors too

PR #281, opened 2020-01-10
Added Twitter Card

### LINKED ISSUE #274: Add Twitter card + other meta
and other meta tags. Use React helmet for this

jreyes88:
Fixes #274

This is my first pull request on someone else's repo so I hope I'm doing this correctly.

I saw Wes' ticket about adding a Twitter card, so I went ahead and did that. The only thing missing is the image for the Twitter card, which I referenced as a file titled "twitter-card.png". I placed this file in `/public`, but that directory is gitignored. It was just a screenshot of the top of the page, but this file could be anything.

I figured `/public` would be a good spot for it since we don't actually need to do any work with it, Twitter just needs a preview image. Attaching it here.

twitter-card

Happy to remove the image line if it's too much trouble.


HugoDF:
@jreyes88 I believe we can use a `static` folder

See https://www.gatsbyjs.org/docs/static-folder/#adding-assets-outside-of-the-module-system

I'm not sure I'm a fan of the last line of tags being cropped in the image you posted.

wesbos:
This is awesome! Could you do one thing and swap out the image with this one?

>>>>>>> master

jreyes88:
I think I got the right image in there. And good call @HugoDF on the `static` folder.

sam-cross:
Better yet, use this image instead!
awesome-uses

jreyes88:
Ack, not sure if I should have closed it. Sorry, I'm new to this. Reopening now.

wesbos:
thanks a ton!

PR #290, opened 2020-01-10
Added john garrett uses

JHGarrett:
Added my uses. Thanks alot Wes!!! You da Man!!!


wesbos:
thanks - it seems this URL isn't a /uses url. Could you fix that?

JHGarrett:
Sorry about that. Should be all good now. You are still the Man!

PR #292, opened 2020-01-10
adds Dieter Stinglhamber

Elhebert:


sam-cross:
The Twitter handle is wrong. I think if you don't have one you're supposed to leave blank

Elhebert:
My bad 😂
Went a bit too fast when adding my information.

PR #294, opened 2020-01-10
Update data.js

alessbell:
Adds info for me, Alessia Bellisario 👋


alessbell:
I see that the linter is failing due to the flag, but I've just used the Canadian flag emoji... any ideas?

wesbos:
someone snuck by - thanks! will fix it

PR #297, opened 2020-01-10
Removed entries which don't link to uses (or similar) pages

sam-cross:
Quickly went through URL entries and removed the people who didn't link to uses pages, and instead to personal portfolios. 

Thankfully there were only three this time!


src/data.js
```diff
@@ -78,6 +78,23 @@ const pages = [
       'Blogger',
     ],
   },
+  {
+  	name: 'Sam Cross',
+  	description: '"Developer", IT student, frequent traveller by coach, addicted to music.',
+  	url: 'https://sam-cross.github.io/uses',
```
iamandrewluca: `404`

sam-cross: Oops - that wasn't supposed to be in the PR. Didn't realise updating the master branch on my repo automatically added commits to this - sorry! I'll try to get that link fixed anyway.

iamandrewluca:
I think we should add a validator that detects 404 pages

wesbos:
thanks! There is an issue for a 404 checker

@sam-cross - thanks for allyour help - I gave you merge access if you want to just merge things :)

PR #318, opened 2020-01-11
Update data.js

conrad96:


src/data.js
```diff
@@ -78,6 +78,25 @@ const pages = [
...
     ],
   },
+  {
+    name: 'Conrad Mugisha',
+    description:
+      'Software Developer. Lover of challenges',
+    url: 'https://aless.co/uses',
+    twitter: '@conradmugisha',
+    emoji: '😎',
+    country: 'UG',
```
palashmon: Please pull your country emoji code from https://github.com/wesbos/awesome-uses/blob/master/scripts/flags.js

src/data.js
```diff
@@ -78,6 +78,25 @@ const pages = [
       'Blogger',
     ],
   },
+  {
+    name: 'Conrad Mugisha',
+    description:
+      'Software Developer. Lover of challenges',
+    url: 'https://aless.co/uses',
```
HugoDF: This is Alessia Bellisario's "uses" page I believe

conrad96: oh snap, forgot to edit that

src/data.js
conrad96: okay. right (y)

src/data.js
```diff
@@ -78,6 +78,25 @@ const pages = [
       'Blogger',
     ],
   },
+  {
+    name: 'Conrad Mugisha',
+    description:
+      'Software Developer. Lover of challenges',
+    url: 'https://github.com/conrad96',
```
HugoDF: This needs to be a "uses" URL, I think we're allowing Gist's containing your equipment & software if you don't have a site where you can add the page.

HugoDF:
@conrad96 closing this for now, feel free to re-open this when you've got a "uses" page (doing it in a Gist is fine)

PR #331, opened 2020-01-11
Add Carlos Sandoval

cdsandoval:


src/data.js
```diff
@@ -2064,6 +2064,25 @@ const pages = [
       'YouTuber',
     ],
   },
+  {
+    name: 'Carlos Sandoval',
+    description: 'Full Stack Developer',
+    url: 'https://carlossandoval.me/',
```
HugoDF: The URL needs to be a "uses" page.

Let me know when that's been fixed.

PR #335, opened 2020-01-12
Use typeface-fira-mono instead of locally checked-in files

KyleAMathews:
package.json all the things!

Cool site!


wesbos:
@HugoDF @KyleAMathews  could we revert this? This was a custom build of Fira Mono that had italics

wesbos:
Oh I can revert right from the UI! Nice : )

PR #338, opened 2020-01-12
Request 50px avatar from clearbit

davidosomething:
would be better to make this source a remote image as a file node so gatsby-image can optimize it


HugoDF:
I just tried it out and it looks good to merge, it's using up about half the image bandwidth as well which is always good

PR #342, opened 2020-01-12
feat: Implement Algolia InstantSearch

Kosai106:
### Intro:
Congratulations on the launch of [uses.tech](https://uses.tech/)! 👏

From all of us at [Algolia](https://algolia.com/), to all of you. ~Merry Christmas and happy holidays.~ Missed the holidays this year.

---

Currently the experience is practically identical to what it is without this PR, however this would allow for easy extensions when it comes to discoverability which could be beneficial when seeing how many people wants to be added to the site.

Next steps could be:

- Search input
- Search highlighting
- multi-facet filtering
- And more

---

~The only thing I'm not entirely happy with in this PR is the way the indexing is working - It's not really an issue, but it works and I doubt this is a project that'll reindex very often. I'll discuss with some colleagues to see if we can come up with something smarter if you'd like.~

Discussed it with some colleagues and overwriting the index on build time is the recommended solution, so all good. 👍


Kosai106:
Please note that the Netlify build breaks because the Algolia admin key has to be added.

iamandrewluca:
Maybe would be better a simple search, without a external service dependency?

delanebob:
> Maybe would be better a simple search, without an external service dependency?

I was thinking the same, instead of using an external service. I've used https://fusejs.io/ in the past and it was very nice to use.

wesbos:
@Kosai106 - how do I get the Admin API key to test this?

Kosai106:
> @Kosai106 - how do I get the Admin API key to test this?

I'll email you.

davidak:
>Currently the experience is practically identical to what it is without this PR, however you will depend on our external, proprietary service

fixed

davidak:
There is no activity since about a year, so i close the PR.

I personally don't like to depend on a proprietary service.

Kosai106:
@davidak There has been no activity as I'm awaiting Wes' response on testing the feature. You're closing this preemptively.

davidak:
He can reopen when he is still interested in this change.

PR #351, opened 2020-01-12
Add-Seyhun-Akyürek (updated data.js)

seyhunak:


src/data.js
```diff
@@ -2064,6 +2064,26 @@ const pages = [
       'YouTuber',
     ],
   },
+  {
+    name: 'Seyhun Akyürek',
+    description:
+      'Senior Software Specialist. Swift, Objective-C, former Rubyist. Author of twitter-bootstrap-rails',
+    url: 'https://www.seyhunakyurek.com/uses/',
```
HugoDF: This URL 404s, let me know when you've addressed this.

seyhunak: Hello,  addressed url is fixed.

Issue #366, opened 2020-01-13
Fix "uses" pages that error


HugoDF:
Outcome of #306, some existing pages weren't compliant, they have now been commented out.

See https://github.com/wesbos/awesome-uses/pull/363/files#diff-be99d06c129c1d2d7d05e3876e275e21R1968-R2239

We should notify the owners of these pages so they can rectify the situation.

## Affected users & pages

- Timed out 
  - Troy Forster, https://tforster.com/uses, @tforster (same on Twitter)
  - Carlos Andres Charris Sandoval, https://carloscharris.com/uses, Twitter @Darkklitos
  - Carlos Junod,  http://carlosjunod.me/uses/, Twitter `@carlosjunod`

- No "uses" page (404)
  - Hussain Aminu, https://hussain4real.github.io/Portfolio/uses, @hussain4real (same Twitter Handle)
  - Sanket Gandhi, https://sanketgandhi.com/uses', Twitter `@sanketgandhi876`
  - Martin Chammah, https://martinchammah.dev/uses, Twitter `@codigomate`
  - Richard Palacios G., https://www.richardpalaciosg.dev/uses, Twitter `@rpalaciosg_`
  - Marek Racík, https://racik.info/uses, Twitter `@marek_racik`
  - Oscar Sánchez, https://devlusaja.com, Twitter `@dev_lusaja`
  - Khriztian Moreno, https://khriztianmoreno.dev/uses, Twitter `@khriztianmoreno`
  - Yuri Yakovlev, https://mynameisyuri.com/uses, Twitter `@yakovlevyuri`
  - "Uses" page is empty - Ændrew Rininsland, https://aendrew.com/uses, Twitter `@aendrew`

- Abuse
  - redirects to a YouTube Video + dodgy certificate - Juan Manuel Incaurgarat, https://kilinkis.me/uses, Twitter: `@kilinkis`

sam-cross:
`Carlos Junod, http://carlosjunod.me/uses/, Twitter @carlosjunod` redirects to homepage. There's no uses info there

wesbos:
@iamandrewluca is going to remove these. Please close once they are gone

iamandrewluca:
Fixed in https://github.com/wesbos/awesome-uses/commit/7a938dd2a68fada76f94c72ce79c17457ff2d2f8#diff-be99d06c129c1d2d7d05e3876e275e21

Run a test on all links from master, all are valid pages

PR #367, opened 2020-01-13
Fix more flakey sites + fake green data check

HugoDF:
Relates to #366


HugoDF:
Merged myself because the data check was exiting with 0 (fine) even when there were errors...

PR #372, opened 2020-01-14
Adding myself, Silvestar Bistrović.

maliMirkec:
Just my setup.

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -35,7 +35,7 @@ const pages = [
       'HTML',
       'CSS',
       'JavaScript',
-      'WordPress'
+      'WordPress',
       'Serverless',
       'JAMstack',
       'Blogger',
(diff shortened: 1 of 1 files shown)


palashmon:
Really nice setup you got there. Thanks for sharing the pic. I think this is the first `/uses` page that I have seen having a setup pic. +1 

P.S. I wish there was a way to +1 the best `/uses` page is the uses.tech site and to show top 5 `/uses` pages based on that rating in a new section somewhere, so that it could inspire others also.

maliMirkec:
@palashmon Thanks for the comment, I appreciate it. 

Also, I think the scoring feature could be an excellent improvement for the site.

HugoDF:
@maliMirkec @palashmon do you want to create an issue with that "rating" suggestion?

PR #379, opened 2020-01-14
build: use unavatar.now.sh for resolving user avatar

### LINKED ISSUE #369: Image graceful fallback
Since the images are pulled from site meta data, it seems "broken" to display broken images.

I'd like to suggest having a more graceful fallback placeholder image and add instructions for what meta data is referenced in the submission instructions.

Kikobeats:
Closes #369

https://unavatar.now.sh uses Clearbit API logo (and more providers) under the hood.

It supports fallback image and it's customizable, e.g., https://unavatar.now.sh/daftpunk?fallback=https://i.imgur.com/0d1TFfQ.jpg


Kikobeats:
@HugoDF the service doesn't offer resize capabilities but it's a thing can be achieve using `images.weserv.nl`:

```
<img src="//images.weserv.nl/?url=https://unavatar.now.sh/daftpunk?fallback=https://i.imgur.com/0d1TFfQ.jpg&w=50">
```

HugoDF:
@Kikobeats let's move discussion to the issue you've created #369

Kikobeats:
@HugoDF updated using resizing 🙂

Kikobeats:
to me it makes sense, since every CDN is doing the right thing, plus avatar doesn't change too often and most of the time the result is served from cache.

HugoDF:
Let's give @wesbos and other maintainers a chance to have their say 😸

Kikobeats:
I added some image optimizations that could be very difficult to achieve without using a CDN 🙂

They are widely tested; we use them for serving +2M requests every month:
https://github.com/microlinkhq/sdk/blob/master/packages/react/src/utils/index.js#L85

Plus, images.weserv.nl is sponsored by CloudFlare ☁️

wesbos:
I say lets do it.

Do we need to resize them? We at least need to use 100x100 for retina screens. The current 50x50 is blurry on my 4k monitor.

I say we just use this and then use the lazy attribute on the images

wesbos:
plus this is better than clearbit because we can try twitter first, and then try their website

wesbos:
merging it - we can change if needed

PR #400, opened 2020-01-15
Remove the first letter(@) from the twitter url

cristianbgp:
Removing the first letter(@) from the twitter url, because it doesn't work on Android devices with the Twitter app.

Example:
From `https://twitter.com/@cristianbgp` to `https://twitter.com/cristianbgp`

### DIFF
--- a/src/components/Person.js
+++ b/src/components/Person.js
@@ -60,7 +60,7 @@ export default function Person({ person, currentTag }) {
         {person.twitter && (
           <TwitterHandle>
             <a
-              href={`https://twitter.com/${person.twitter}`}
+              href={`https://twitter.com/${person.twitter.replace('@', '')}`}
               target="_blank"
               rel="noopener noreferrer"
             >


src/components/Person.js
```diff
@@ -60,7 +60,7 @@ export default function Person({ person, currentTag }) {
         {person.twitter && (
           <TwitterHandle>
             <a
-              href={`https://twitter.com/${person.twitter}`}
+              href={`https://twitter.com/${person.twitter.slice(1)}`}
```
iamandrewluca: Please make a check first if twitter starts with `@` something like, otherwise it will strip first letter on usernames that were added without `@`
```js
person.twitter.startsWith('@') ? person.twitter.slice(1) : person.twitter
```

HugoDF: Or even
`.replace('@', '')`

iamandrewluca: With replace will replace even in the middle of the name `some@handle` -> `somehandle`

HugoDF: Twitter handles can't have `@` in them
As per Twitter Help Center:
> A username can only contain alphanumeric characters (letters A-Z, numbers 0-9) with the exception of underscores, as noted above. Check to make sure your desired username doesn't contain any symbols, dashes, or spaces.

iamandrewluca: Ok, agree, `.replace('@', '')` can be used then

cristianbgp: I agree, let's use replace. Also `replace` is used to show the actual twitter username in line [68](https://github.com/wesbos/awesome-uses/pull/400/files#diff-8cb77f435c7ded7519d20549e532dc6cL68)

PR #416, opened 2020-01-15
adding uses page to data.js

shan5742:


HugoDF:
As per discussion, re-open this with just your new changes please 🙏

PR #406, opened 2020-01-15
feat: add precommit hook to lint fix data.js

iamandrewluca:

PR #422, opened 2020-01-16
feat: change validate action to take into consideration only added user

### LINKED ISSUE #382: Adapt validation action to take into consideration only added persons

iamandrewluca:
`node scripts/data-valdiate.js` cand also be run locally

This also will fix the issue when some valid URL returned timeout
Because the action wanted to make all requests at the same time

`git restore` is available only from Git 2.23

https://github.blog/2019-08-16-highlights-from-git-2-23/#experimental-alternatives-for-git-checkout

Also added cache action to speed up npm install between action executions.

Closes #382

### DIFF
--- a/.github/workflows/data-validate.yml
+++ b/.github/workflows/data-validate.yml
@@ -15,11 +15,15 @@ jobs:
     - uses: actions/setup-node@v1
       with:
         node-version: 13.x
-    - name: Install validation libs
-      run: |
-        npm install -g @hapi/joi@17.0.2
-        npm install -g @actions/core@1.2.0
-        npm link @hapi/joi
-        npm link @actions/core
+
+    - name: Cache/Restore node modules
+      uses: actions/cache@v1
+      with:
+        path: ~/.npm # npm cache files are stored in `~/.npm` on Linux/macOS
+        key: ${{ runner.os }}-npm-${{ hashFiles('./package-lock.json') }}
+
+    - name: Install Dependencies
+      run: npm install
+
     - name: Validate data.js
       run: node ./scripts/data-validate.js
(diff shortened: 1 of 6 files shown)


wesbos:
You are good! I was just thinking about how to do this, and how to diff them. Great idea! will try it out in a bit :)

.github/workflows/data-validate.yml
```diff
@@ -21,5 +21,10 @@ jobs:
         npm install -g @actions/core@1.2.0
         npm link @hapi/joi
         npm link @actions/core
+        mv src/data.js src/tmp-data.js
```
HugoDF: I would be tempted to move all the `mv` & shell/git stuff into a `scripts/data-validate.sh` so that it can be run locally as well as in the GitHub action

Contents would be:

```sh
mv src/data.js src/tmp-data.js
git fetch origin master
git restore --source=FETCH_HEAD src/data.js
mv src/data.js src/master-data.js
mv src/tmp-data.js src/data.js
node ./scripts/data-validate.js
```

iamandrewluca: I also will add a check if is run on master will run for all users.

iamandrewluca:
@wesbos @HugoDF take a look now.

HugoDF:
For @wesbos or anyone else trying to pull locally, you need git v2.25 for git-restore to exist, on Mac, that means installing/upgrading from 2.21 (brew install git & restart terminal session).

iamandrewluca:
Yes action will be run only on PR, but you can run the validate script event locally. `node scripts/data-valdiate.js`

HugoDF:
@iamandrewluca fair enough, I think I must have confused myself,

I think we mentioned running against master on a schedule right? Guessing that's better done as a follow-up PR.

iamandrewluca:
I think you misunderstood this PR with this one #406

PR #424, opened 2020-01-16
Add Will Presley - update data.js

WillPresley:
Love this idea! Looking forward to watching it grow and hopefully helping out with this specific project!


HugoDF:
Welcome to uses.tech @WillPresley

PR #437, opened 2020-01-17
style: fix eslint issues

HugoDF:
Fix most eslint issues.


iamandrewluca:
As a note, commits that format the code, better to be named something like this
```
style(module-if-any): some description
```

https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#type

PR #443, opened 2020-01-17
fix(tags): use lowercase tags to avoid duplication

### LINKED ISSUE #441: Javascript vs JavaScript
Could maybe normalize tags and prefer the most common one?

iamandrewluca:
Closes #441


iamandrewluca:
Also strange that when selecting a text filter, 0 person remains.

wesbos:
I'd like to maintain case on the tags - so I'm gonna say we just fix data.js instead. 

Maybe once things calm down, we can take a look at the tags and then implement an ENUM of available tags..

PR #473, opened 2020-01-20
chore: downgrade scripts support to node 12

iamandrewluca:
From start scripts needed to be run with node 13 (esm support)
Because data.js was a esm export, and needed to be compiled
Moved all scripts to commontjs, and data.js file
So can be imported by scripts without compiling


HugoDF:
@iamandrewluca Instead of doing this, couldn't we have leveraged `esm` (`node -r esm path-to-script.js`)?

iamandrewluca:
Let's keep them as commonjs, because esm is now experimental.

PR #475, opened 2020-01-20
feat: add HTML-only back to top link

### LINKED ISSUE #426: Feature: back to top button
As the list grows, it's too long to scroll back up to top. I have already developed the code.

```js
const backToTop = document.querySelector('back-to-top')
window.onscroll = function () {
  scrollFunction()
}
backToTop.addEventListener('click', backToTopFunction)

function scrollFunction() {
  if (document.body.scrollTop > 20 || document.documentElement.scrollTop > 20) {
    backToTop.style.display = 'inline-block'
  } else {
    backToTop.style.display = 'none'
  }
}

function backToTopFunction() {
  document.body.scrollTop = 0
  document.documentElement.scrollTop = 0
}
```
```css
.backToTop {
    transform: rotate(90deg);
    position: fixed;
    bottom: 1%;
    right: 1%;
    background: var(--pink);
    font-size: 3rem;
    cursor: pointer;
    border-radius: 3px;
    padding: 1rem;
    transition: background-color 0.2s ease 0s;
    display: none;
}
```
```html
<div class="back-to-top">‹</div>
```

HugoDF:
Potentially, closes #426 
- add "top" id to main `h1`
- add a fixed-position link to `#top`

## Code Modifications

* Ensure the code submitted is formatted similarly to existing code
* Ensure variable, method, function, and component names are clear and concise

## Final Requirements

**Add your name to verify:** Hugo

### DIFF
--- a/src/components/header.js
+++ b/src/components/header.js
@@ -21,7 +21,7 @@ function Header({ siteTitle, siteDescription, siteUrl }) {
         <meta name="twitter:image" content={`${siteUrl}/twitter-card.png`} />
       </Helmet>
       <div>
-        <h1>
+        <h1 id="top">
           <Link to="/">/uses</Link>
         </h1>
         <p>
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -7,6 +7,12 @@ import Layout from '../components/layout';
 import Person from '../components/Person';
 import Topics from '../components/Topics';
 
+const BackToTopLink = styled.a`
+  position: fixed;
+  bottom: 10px;
+  right: 10px;
+`;
+
 function IndexPage() {
   const { currentTag } = useContext(FilterContext);
   const { allPerson } = useStaticQuery(graphql`
@@ -43,6 +49,7 @@ function IndexPage() {
           <Person key={person.name} person={person} currentTag={currentTag} />
         ))}
       </People>
+      <BackToTopLink href="#top">&uarr; Back to top</BackToTopLink>
     </Layout>
   );
 }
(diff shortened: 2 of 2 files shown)


HugoDF:
cc @iamandrewluca @vikasbhvsr

joshmanders:
It's a good start, simple and works, but it kind of ugly. Maybe put a background around the link and make the background slightly transparent so it overlays everything but doesn't look hideous. 

before: 
image

after:
image

css:
```css
background: rgba(31, 31, 31, 0.8);
padding: 1rem;
border-radius: .8rem;
```

HugoDF:
@joshmanders thanks for the feedback 👍

vikasbhvsr:
> @joshmanders thanks for the feedback 👍
`
.backToTop {
    position: fixed;
    bottom: 1%;
    right: 1%;
    background: var(--pink);
    font-size: 3rem;
    cursor: pointer;
    border-radius: 3px;
    padding: 1rem;
    transition: background-color 0.2s ease 0s;
    display: none;
}`

You can use my styling.

HugoDF:
@vikasbhvsr updated to match, dropped the large font because the link does appear even when you're at the top of the page

we can have a look at hiding it based on scroll position later

sam-cross:
I think it should be hidden when the user is scrolled right to the top. A simple `display: none;` would suffice, or better yet, throw it off-page by default and bring it in when the user has scrolled down any amount.

Also, how's it looking on mobile? Is it responsive?

wesbos:
I just merged this - will see how it looks and change it up if need be

HugoDF:
@sam-cross should work ok on mobile but would be nice to have hiding based on scroll position

PR #505, opened 2020-01-21
Add @asurbernardo

asurbernardo:
Asur Bernardo

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -94,6 +94,27 @@ module.exports = [
     ],
   },
   {
+    name: 'Asur Bernardo',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -93,6 +93,27 @@ module.exports = [
       'Maker',
     ],
   },
+  {
+    name: 'Asur Bernardo',
+    description: "Back end developer with no aesthetic sense. Full stack with reservations. Open-source enthusiast. Continuous learner!',
```
HugoDF: You probably want to do:
```suggestion
    description: 'Back end developer with no aesthetic sense. Full stack with reservations. Open-source enthusiast. Continuous learner!',
```

asurbernardo: Whooops! That was dumb, thanks for the catch! :wink:

PR #558, opened 2020-01-26
Add @Zoran_Panev

ZokiPokiDev:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -2566,7 +2566,7 @@ module.exports = [
     emoji: '🛡',
     country: '🇲🇰',
     computer: 'linux',
-    phone: 'f-droid',
+    phone: 'android',
     tags: [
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -2558,6 +2558,27 @@ module.exports = [
       'GraphQL',
     ],
   },
+  {
+    name: 'Zoran Panev',
+    description: 'Web developer',
+    url: 'https://gist.github.com/8f08557764711afbf82b75ac0ce61e79.git',
```
HugoDF: You're gonna need to actually put the gear/software you use in this gist

ZokiPokiDev: > You're gonna need to actually put the gear/software you use in this gist

Did you mean something like this? Is it now fine? I have updated my gist.

PR #559, opened 2020-01-27
trigger change so picks up new twitter avatar

markhorsell:


iamandrewluca:
This will not solve the problem.
As you can see at this URL, unavatar cannot find any related image.

https://images.weserv.nl/?url=https://unavatar.now.sh/markhorsell.com&w=100&l=9&af&il&n=-1

PR #561, opened 2020-01-27
add scroll to top compatibility with other browsers

saleebm:
Hey man, just noticed the scroll to top doesn't show on other browsers besides Firefox. I gotta love listening to SyntaxFM so I know you were only using Firefox for a month straight.

I really like this site, I just figured to contribute. Thank you!

I found the function scrollTopMax here - https://stackoverflow.com/a/36368796

### DIFF
--- a/src/components/BackToTop.js
+++ b/src/components/BackToTop.js
@@ -24,6 +24,14 @@ const BackToTopLink = styled.a`
   }
 `;
 
+function scrollTopMax() {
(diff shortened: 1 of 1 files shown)


wesbos:
thanks a ton! just simplified it a bit more here: https://github.com/wesbos/awesome-uses/commit/4c0cb690d3a27a3b995b627d085da411e4935b87

PR #565, opened 2020-01-28
Add @ahmed_ibrahhim

AhmadIbrahiim:


src/data.js
```diff
@@ -22,6 +22,27 @@ module.exports = [
       'CSS',
     ],
   },
+  {
+    name: 'Ahmed Ibrahim',
+    description:
+      'Full-stack developer who believes that Passion is the key to every success.',
+    url: 'https://ahmed-ibrahim.com/?/uses/',
```
HugoDF: This isn't a uses page

PR #571, opened 2020-01-30
Add Jenna Pederson

jennapederson:
Add Jenna Pederson

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -2401,7 +2401,7 @@ module.exports = [
   {
     name: 'Jenna Pederson',
     description:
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -2398,6 +2398,29 @@ module.exports = [
       'Blogger',
     ],
   },
+  {
+    name: 'Jenna Pederson',
+    description:
+      'technical entrepreneur with a passion for community and equity-building',
```
sam-cross: A little nitpicky, but could you make this start with an uppercase T? Though the checks are "failing", nothing is actually wrong with this, so once that's done we can merge. Thanks!

PR #608, opened 2020-02-06
Added @jamieryandev

MrJamieRyan:


MrJamieRyan:
Added wrong country flag

PR #613, opened 2020-02-06
Add to data shaunfarrell

codedplanetuk:


src/data.js
```diff
@@ -292,6 +292,31 @@ module.exports = [
...
+      'Developer',
+      'Web Developer',
+      'Full Stack',
+      'Front End',
+      'JavaScript',
+      'TypeScript',
+      'React',
+      'HTML',
+      'CSS',
+      'ReactNative',
```
sam-cross: Please put a space in React Native to keep it consistent with other entries - thank you!

PR #615, opened 2020-02-07
Add to data tcgumus

tcgumus:
Thanks

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -137,6 +137,30 @@ module.exports = [
     ],
   },
   {
+    name: 'Tuna Çağlar Gümüş',
+    description:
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -136,6 +136,30 @@ module.exports = [
...
+    twitter: '@tc_gumus',
+    emoji: '🧿',
+    country: '🇹🇷',
+    computer: 'apple',
+    phone: 'iphone',
+    tags: [
+      'Developer',
+      'Full Stack',
+      'JavaScript',
+      'JAVA',
```
sam-cross: Could you make "JAVA" -> "Java" to stay consistent with other entries - thank you!

tcgumus: Got it. Thank you 👍

PR #660, opened 2020-02-20
Adding Colin Principe colin.principe@gmail.com

FatherOfCurses:


src/data.js
```diff
@@ -332,6 +332,34 @@ module.exports = [
       'Gatsby',
     ],
   },
+  {
+    name: 'Colin Principe',
+    description:
+      'Software Guild graduate, Full stack developer, Java backend, Angular/React frontend.',
+    url: 'http://principe.io/uses',
```
HugoDF: This isn't a uses page

FatherOfCurses: URL changed, was having issues with my hosting.  Latest commit resolves properly.
[https://github.com/FatherOfCurses/awesome-uses/commit/9d2d15026f52119d3335fbfcca759cad6945ca75](Latest commit)

src/uses.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
HugoDF: Could we remove this file?

FatherOfCurses: Deleted from my commit.

FatherOfCurses:
Going to cancel this pull request and submit a new one because checks are failing.

PR #723, opened 2020-03-29
Add Ed Henderson

edhenderson:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -1315,6 +1315,7 @@ module.exports = [
   {
     name: 'Ed Henderson',
     description: 'Unicorn amongst donkeys',
+    url: 'https://twitter.com/edhenderson',
     twitter: '@edhenderson',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -1312,6 +1312,27 @@ module.exports = [
       'Photographer',
     ],
   },
+  {
+    name: 'Ed Henderson',
+    description: 'Unicorn amongst donkeys',
+    url: 'https://twitter.com/edhenderson',
```
gatarelib, line 1318: the URL has to be a ` /uses`

edhenderson: I actually hadn't even put a URL in originally but it insisted on one.

iamandrewluca:
`awesome-uses` is a place for people to share their environment and what software/hardware they are using.

Unfortunately provided URL is not a uses page. Fell free to open a new PR with the right details.

PR #732, opened 2020-04-03
Add Jesús Leganés-Combarro 'piranna'

piranna:
Jesús Leganeś-Combarro 'piranna'

I've added myself at end, but this is a random place too :-)

Offtopic: `data.js` file is really big and difficult to maintain, I think it would be nice to use instead a folder with one file for each user, what do you think?


piranna:
Thank you :-)

PR #745, opened 2020-04-16
Add Pedro Sousa

pedrosousaef:
Pedro Sousa


iamandrewluca:
`awesome-uses` is a place for people to share their environment and what software/hardware they are using.

Unfortunately provided URL is not a uses page. Fell free to open a new PR with the right details.

PR #763, opened 2020-04-24
Add Karel De Smet

carlos-ds:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -3998,6 +3998,28 @@ module.exports = [
     ],
   },
   {
+    name: 'Karel De Smet',
+    description:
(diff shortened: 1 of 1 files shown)


carlos-ds:
> Could we try this again without changing unrelated entries in the data file?

Sorry Hugo and thank you for the review. I'm new to this so still trying to find my way. I believe prettier was the cause of all the unwanted changes.

I undid my changes by doing a git reset on my branch to the commit before my changes. And then I made my changes again, with only the insertion of my data.  I hope that's sufficient but please let me know if it's not.

HugoDF:
Nice one let me double check it and merge it hopefully.

PR #764, opened 2020-04-24
Update data.js

ErikThiart:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -23,6 +23,25 @@ module.exports = [
     ],
   },
   {
+    name: 'Erik Thiart',
+    description:
+      'Full Stack Developer, I am a Swiss Army knife.',
+    url: 'https://erikthiart.com/uses/',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -22,6 +22,25 @@ module.exports = [
       'Platforms',
     ],
   },
+  {
+    name: 'Erik Thiart',
+    description:
+      'Full Stack Developer, I am a Swiss Army knife.',
+    url: 'https://erikthiart.com/uses/',
```
HugoDF, line 29: This just redirects to the homepage, could you create a uses page?

ErikThiart: Oopsie... I thought it was just robots adding us... Hehe, caught red-handed.

I have added a uses page https://erikthiart.com/uses/index.php and removed the .htaccess page, check if it works for you now?

HugoDF: Cool works now

PR #768, opened 2020-04-24
Update data.js with Keith Donegan

MadebyToast:
Adding myself to this (hopefully)!

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -1222,7 +1222,7 @@ module.exports = [
     url: 'https://www.keithdonegan.com/uses/',
     twitter: '@eirestudio',
     emoji: '🤪',
-    country: '🇮🇪🇸',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -1215,6 +1215,23 @@ module.exports = [
...
     ],
   },
+  {
+    name: 'Keith Donegan',
+    description:
+      'Irish WordPress developer, based in London, UK.',
+    url: 'https://www.keithdonegan.com/uses/',
+    twitter: '@eirestudio',
+    emoji: '🤪',
+    country: '🇮🇪🇸',
```
HugoDF, line 1225: Any reason there's 2 emoji? CI won't pass until this field contains a single flag

MadebyToast: No idea, I'll change.

PR #778, opened 2020-04-28
adding myself @roniemeque 😁

roniemeque:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -3589,6 +3589,27 @@ module.exports = [
     tags: ['JavaScript', 'TypeScript', 'React', 'Node', 'GraphQl', 'Svelte'],
   },
   {
+    name: 'Ronie Meque',
+    description:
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -3588,6 +3588,27 @@ module.exports = [
     phone: 'android',
     tags: ['JavaScript', 'TypeScript', 'React', 'Node', 'GraphQl', 'Svelte'],
   },
+  {
+    name: 'Ronie Meque',
+    description:
+      'Web Developer. Loves to help people and hopes to one day actually finish a side project or travel the world, whichever comes firts',
+    url: 'https://ronie.dev/uses',
```
HugoDF, line 3595: Hmm this doesn't look like a uses page to me, would it be cool if we hold off until you create one?

roniemeque: You're right, sorry! I will finish it and notify here 😊

iamandrewluca:
Fell free to open a new PR when your page will represent a `/uses` page. 🙂

PR #784, opened 2020-04-29
Add profile to data.js

SGolbert:


src/data.js
```diff
@@ -1,5 +1,28 @@
 // keep it commonjs export
 module.exports = [
+  {
+    name: 'Sebastian Golbert',
+    description:
+      'Full-stack Developer. Chemist. Data Scientist',
+    url: 'https://github.com/SGolbert',
```
HugoDF, line 7: This needs to be a "uses" page, a GitHub gist is fine too

PR #843, opened 2020-05-30
Add Jay Tyrrell

jaytyrrell13:
Jay Tyrrell

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -242,6 +242,17 @@ module.exports = [
     ],
   },
   {
+    name: 'Jay Tyrrell',
+    description: 'Full Stack Developer',
+    url: 'https://jaytyrrell.co/uses/',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -241,6 +241,17 @@ module.exports = [
       'Teacher',
     ],
   },
+  {
+    name: 'Jay Tyrrell',
+    description: 'Full Stack Developer',
+    url: 'https://jaytyrrell.co/uses/',
```
gatarelib, line 247: Is it my internet or is your /uses page empty?

jaytyrrell13: That's strange. My /uses page is filled out. Could you try again? Are you trying with https?

HugoDF: I've checked it, it's filled out

PR #857, opened 2020-06-19
feat: Clean up tags 

okcoker:
This normalizes and cleans up the tags on the page in 2 ways: 
- Merges "case-varying" tags like "JavaScript" and "Javascript" to be merged into whatever tag is the most popular
- Adds a manual transformation for things that aren't necessarily case-varying like "Full Stack" vs "Fullstack" or "React" vs "React.js". I've manually added the replacement rules [here](https://github.com/wesbos/awesome-uses/compare/master...okcoker:feature/tag-normalization?expand=1#diff-9af8e32cad39119264076f2a7289ae6eR18-R25) based on the current count live on the site but it can be altered as you see fit (or not at all)

This means people can still enter a tag like "Javascript" and it will be changed to "JavaScript" (not because it's correct, but because it's the most popular) at build time.


wesbos:
thanks a ton! super helpful

PR #877, opened 2020-07-03
add-phemmy

adexphem:


src/data.js
```diff
@@ -8855,6 +8855,26 @@ module.exports = [
       'WordPress',
     ],
   },
+  {
+    name: 'Oluwafemi Adeosun',
+    description: '💻Frontend Engineer and Politician based in Lagos, Nigeria.',
+    url: 'https://github.com/adexphem',
```
HugoDF, line 8861: this needs to be a uses page, a gist or repo is fine

davidak (changes requested):
what HugoDF said

PR #891, opened 2020-07-16
Add Adam Siekierski

AdamSiekierski:
Let's add another `uses` page :). Maybe someday, someone will do something nice with all that data :D. Maybe some statistics or data analysis - it looks like a nice dataset.

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -420,7 +420,7 @@ module.exports = [
   {
     name: 'Adam Siekierski',
     description:
-      'Non-proffessional fullstack developer. Big fan of JavaScript, TypeScript, React and Vue. Co-host of Polish webdevelopment podcast.',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -417,6 +417,30 @@ module.exports = [
       'Teacher',
     ],
   },
+  {
+    name: 'Adam Siekierski',
+    description:
+      'Non-proffessional fullstack developer. Big fan of JavaScript, TypeScript, React and Vue. Co-host of Polish webdevelopment podcast.',
```
xxczaki, line 423:
```suggestion
      'Non-professional fullstack developer. Big fan of JavaScript, TypeScript, React, and Vue. Co-host of Polish web development podcast.',
```

PR #896, opened 2020-07-20
Add Vyshnav Gangadharan

Vyshnav2255:


src/data.js
```diff
@@ -81,6 +81,30 @@ module.exports = [
       'Photoshop',
     ],
   },
+  {
+    name: 'Vyshnav Gangadharan',
+    description: 'An engineer who loves Graphics Designing, Film making, Video editing & Coding.',
+    url: 'https://vyshnav.netlify.app/more/',
```
HugoDF, line 87: In order for this to pass validation, the URL needs to contain `uses` and be a `uses` page (although I've seen that you've got your uses information on Dropbox).

davidak, line 87: 404 Page not Found!

should be uses page

davidak (changes requested):

PR #903, opened 2020-07-26
Added David Amunga

DavidAmunga:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -1,6 +1,30 @@
 // keep it commonjs export
 module.exports = [
   {
+    name: 'David Amunga',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -1,5 +1,29 @@
 // keep it commonjs export
 module.exports = [
+  {
+    name: 'David Amunga',
+    description: '24 | Software Engineer who loves building awesome products',
+    url: 'https://github.com/DavidAmunga/awesome-uses/',
```
HugoDF, line 6: This should be a uses page/repo not a fork of this repository

davidak (changes requested):
what HugoDF said

PR #946, opened 2020-09-22
Update Scott Spence Uses URLs

spences10:
Updating my Uses URLs the current one points to an old domain, I've added a short link so I don't have to submit another PR is the URL changes again 👍


davidak:
You only need to update `src/data.js`.

davidak (changes requested):
You only need to update `src/data.js`.
The other files are generated automatically.

spences10:
Cool! Reverted those files

davidak:
I still see changes to the `readme.md`.

https://github.com/wesbos/awesome-uses/pull/946/files

Issue #637, opened 2020-02-11
Merging tags to most common spellings


maxpou:
hiya,

I just noticed that same filters have been repeated twice or more.

We have: 
* "Front End", "Front end" and "Frontend"
* "Ruby On Rail", "Ruby on Rails" and "Rails" (rails === RoR right?)
* "Gatsby" and "GatsbyJs"
* "Vue" and "VueJS"

Do you mind if I merge people's tags to the most used?

HugoDF:
@maxpou go for it

There was an issue kicking around to deduplicate tags in the code but the agreed upon approach is to edit the data file.

Ideally, make 1 PR per tag that you're unifying.

maxpou:
@HugoDF there you go. As said before, I arbitrary took the most popular keyword
* "Front End", "Front end" and "Frontend" ==> #639
* "Gatsby" and "GatsbyJs" ==> #640 
* "Vue" and "VueJS"  ==> #641

Someone can confirm this? 
* "Ruby On Rail", "Ruby on Rails" and "Rails" (rails === RoR right?)

HugoDF:
> "Ruby On Rail", "Ruby on Rails" and "Rails" (rails === RoR right?)

This is correct, in this case I would use "Ruby on Rails"

sam-cross:
We could list these common spellings at the top of `data.js` so people can easily reference them.

nico-bachner:
Hi,

I've been cleaning up some tags and realised that since it's actually really easy that this issue should perhaps be labelled "good for first-time" or similar.

Would a maintainer be so kind as to do this if they agree?
Cheers

nico-bachner:
I'm planning on merging the VSCode tags next. 

Which spelling is preferrable: Visual Studio Code, VS Code or VSCode ?

carbontwelve:
I can see that this issue has a long history and @BlakeCampbells looks to have made progress in merging tags. Is this issue now completed, or is there work still to do?

BlakeCampbells:
Good call @carbontwelve , I think this is okay to close. 

To keep it from happening in the future we should add either some validation around tags or in `src/util/stats.js` we should make sure `normalizeTags` covers more cases.

PR #962, opened 2020-10-03
Merge duplicate Node.js tags (Issue #637)

nico-bachner:


src/data.js
```diff
@@ -2427,7 +2427,7 @@ module.exports = [
       'Developer',
       'Software Engineer',
       'Full Stack',
-      'nodejs',
+      'node',
```
davidak, line 2430:
```suggestion
      'Node',
```
that would be a duplicate again

davidak (approved):
looks good

only one typo

HugoDF (changes requested):
Technically "Node.js" is the official spelling https://nodejs.org/en/

nico-bachner:
I used Node because it was the more popular name, but if it's ok, I'd be more than happy to change it to Node.js

HugoDF:
> I used Node because it was the more popular name, but if it's ok, I'd be more than happy to change it to Node.js

I'm not particularly fussy, but if the trademark is on Node.js then we should probably use that 

Also, I think this branch has conflicts that you'll need to resolve.

nico-bachner:
Should be all good now

davidak (approved):

PR #965, opened 2020-10-03
Add Rowe Morehouse

rowe-morehouse:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -1,6 +1,30 @@
 // keep it commonjs export
 module.exports = [
   {
+    name: 'Rowe Morehouse',
+    description:
+      'Growth · Product · Software Project Management · Frontend Dev · Design · Technical Writing · Sales',
+    url: 'https://rowe-morehouse.github.io/resume/uses/',
+    twitter: '@rowemore',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -1,5 +1,29 @@
 // keep it commonjs export
 module.exports = [
+  {
+    name: 'Rowe Morehouse',
+    description:
+      'Growth · Product · Software Project Management · Frontend Dev · Design · Technical Writing · Sales',
+    url: 'https://rowe-morehouse.github.io/resume/uses/',
```
davidak, line 7: the page don't shows what tools you use, so it don't fit's this projecs

davidak (changes requested):

src/data.js
rowe-morehouse: Look again, the page has all the tools I use: [https://rowe-morehouse.github.io/resume/uses/](https://rowe-morehouse.github.io/resume/uses/)

davidak: nice

davidak (approved):
LGTM

HugoDF (approved):

PR #966, opened 2020-10-03
Merge duplicate Web Dev tags (Issue #637)

nico-bachner:
I went with Web Dev since it encompasses both Web Developer and Web Development


davidak (changes requested):
i'm not sure about that

people might not want to be called that

nico-bachner:
Makes sense. What would you suggest instead? Just Web?

PR #969, opened 2020-10-04
Update data.js

DhananjayPorwal:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -3668,6 +3668,27 @@ module.exports = [
     ],
   },
   {
+    name: 'Dhananjay Porwal',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -3667,6 +3667,27 @@ module.exports = [
       'Musician',
     ],
   },
+  {
+    name: 'Dhananjay Porwal',
+    description: 'Cyber Security Analyst',
+    url: 'https://dhananjayporwal.me/',
```
davidak, line 3673: nice website, but not a uses page

so it don't fit this project

davidak (changes requested):

PR #972, opened 2020-10-05
Add Adi Purnomo

medival:


medival (commented):
Update medival change on "computer" and add some tags

HugoDF (changes requested):
I won't be able to merge this if we'be got 10k changes, could you add just yourself to the data.js without changing anything else?

medival:
I'll close this PR, then I create new one https://github.com/wesbos/awesome-uses/pull/984

PR #983, opened 2020-10-07
Added my uses

abba-m:


src/data.js
```diff
@@ -126,6 +126,26 @@ module.exports = [
       'Blogger',
     ],
   },
+  {
+    name: 'Muhammad Mahmood',
+    description: 'Web Dev',
+    url: 'https://www.about.me/abba-m/uses',
```
HugoDF, line 132: this is not a uses page 😄

HugoDF (changes requested):
let us know when you've created a "uses" page not just a URL that redirects to the homepage

abba-m:
Kill me already

PR #987, opened 2020-10-08
Ignore duplicated tags

### LINKED ISSUE #986: Duplicated tags
Hello! I've noticed that some people have duplicated tags (e.g. Fidalgo has Front End twice). This is causing "key prop" warnings from React when rendering the tags list in the Person component and the topics count is also affected by this showing inaccurate numbers.

devrasec:
I've noticed that some people have duplicated tags (e.g. Fidalgo has Front End twice). This is causing "key prop" warnings from React when rendering the tags list in the Person component and the topics count is also affected by this showing inaccurate numbers.

This PR solves this issue.

Closes #986


wesbos:
thanks!

PR #1003, opened 2020-10-15
Add Luis Juarez

IAmLuisJ:


davidak (changes requested):
can you please not have 10,246 additions, 10,205 deletions in this PR? that don't look right

it might already solve it to squash the commits

PR #1004, opened 2020-10-16
Adding Jorge Ruvalcaba

jorgeruvalcaba:
Also removed a duplicated tag I found while scrolling


src/data.js
```diff
@@ -5824,6 +5823,28 @@ module.exports = [
     phone: 'iphone',
     tags: ['Full Stack', 'Developer', 'Front End', 'JavaScript', 'React'],
   },
+  {
+    name: 'Jorge Ruvalcaba',
+    description:
+      'Software Engineer, Senior Frontend Engineer, Taco Master and internet weirdo 🤓',
+    url: 'https://jorgeruvalcaba.dev/uses/',
```
davidak, line 5830: i only get

> Not Found
> You just hit a route that doesn't exist... the sadness.

davidak (changes requested):
LGTM

except that the uses page does not exist... but that is the whole purpose of this

src/data.js
jorgeruvalcaba: My bad 🤦  . Didn't push the page before but now it's alive. You can check it out [here](https://jorgeruvalcaba.dev/uses/).

jorgeruvalcaba:
@davidak resolved! 😅

davidak (approved):

davidak:
very good. thank you!

PR #1013, opened 2020-10-21
Add @arcatech

ArcaTech:
**Add your name to verify:** Dwayne Harris @arcatech


davidak (approved):
Thanks for your contribution!

PR #1030, opened 2020-11-19
Add Maxim Villivald

villivald:
Add Maxim Villivald

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -60,6 +60,26 @@ module.exports = [
     ],
   },
   {
+    name: 'Maxim Villivald',
+    description: 'Web Developer, IT Student 👨‍💻, Blogger & part time Tram Driver 🚃.',
+    url: 'https://proj.ninja/uses',
+    twitter: '@crapp_blog',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -59,6 +59,26 @@ module.exports = [
       'Python',
     ],
   },
+  {
+    name: 'Maxim Villivald',
+    description: 'Web Developer, IT Student 👨‍💻, Blogger & part time Tram Driver 🚃.',
+    url: 'https://proj.ninja/uses',
```
davidak, line 65: The page don't exist

Not Found
The requested URL /uses was not found on this server.

davidak (changes requested):

src/data.js
villivald: I am sorry, there was a strange issue with .htaccess on the server-side. Now it works.

davidak: great, thanks.

i like the individual style of the page

davidak (approved):

Issue #999, opened 2020-10-13
Validate data.js using DangerJS


davidak:
https://github.com/wesbos/awesome-uses/pull/998/checks
```
Run npx danger ci
You need to add GITHUB_TOKEN to your Danger action in the workflow:
  
    - name: Danger JS
      uses: danger/danger-js@X.Y.Z
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
    
The DANGER_GITHUB_API_TOKEN/DANGER_BITBUCKETSERVER_HOST/DANGER_GITLAB_API_TOKEN environmental variable is missing
Without an api token, danger will be unable to comment on a PR
Error:  Error: Cannot use authenticated API requests.
    at Object.getPlatformForEnv (/home/runner/work/awesome-uses/awesome-uses/node_modules/danger/distribution/platforms/platform.js:76:11)
    at Object.<anonymous> (/home/runner/work/awesome-uses/awesome-uses/node_modules/danger/distribution/commands/ci/runner.js:81:57)
    at step (/home/runner/work/awesome-uses/awesome-uses/node_modules/danger/distribution/commands/ci/runner.js:32:23)
    at Object.next (/home/runner/work/awesome-uses/awesome-uses/node_modules/danger/distribution/commands/ci/runner.js:13:53)
    at fulfilled (/home/runner/work/awesome-uses/awesome-uses/node_modules/danger/distribution/commands/ci/runner.js:4:58)
Error: Process completed with exit code 1.
```

Joshua-rose:
Received the same with #1000

jorgeruvalcaba:
I'm getting the same in #1004

HugoDF:
Yeah this is one for @wesbos he needs to give dangerJS the right permissions

rohit-gohri:
I mentioned this in the original issue https://github.com/wesbos/awesome-uses/issues/478#issuecomment-623157116 for automating checks. But forgot to mention in the PR (#796) that a secret `DANGER_GITHUB_API_TOKEN` will need to be added to the project and now it's merged without access to the repo 🙈.

itsmanojb:
Still failing. 

![Screenshot from 2020-10-28 02-25-42](https://user-images.githubusercontent.com/11155266/97361143-25fe9f80-18c5-11eb-87dd-1b0c9c0cdd2e.png)

Issue #478, opened 2020-01-20
Feature: comment on PR with validation output


HugoDF:
It seems possible to leave a comment on PRs when validation fails.

See  https://github.com/unsplash/comment-on-pr, code is obviously Ruby but equivalent functionality in JS should do the trick.

sam-cross:
This would be SO helpful. Might look into this tonight.

HugoDF:
Even better would be to take a screenshot/photo of the uses page being submitted so we don't have to open the URLs to review that it's not a placeholder.

wesbos:
👍🏻 would love this feature for commenting what went wrong

HugoDF:
I'm happy to have a stab at this later today.

sam-cross:
Please do - I tried looking into it but I have no idea how to use GitHub Actions. I did find [this](https://github.com/thollander/actions-comment-pull-request/blob/master/src/main.ts) which might be useful!

HugoDF:
Nice one, I think using pageres + the examples in here we should be able to get this over the line.

HugoDF:
@wesbos @sam-cross PR is up and working

I might publish it as a standalone GitHub action "Leave PR comment"

HugoDF:
I believe since we're integrating more tightly with GitHub, we could also validate that the PR name is formatted as expected & that the PR template has been filled out.

HugoDF:
Interesting, the commenting doesn't  work, probably due to user permissions. We might need to generate another token & use that instead of the default one.

sam-cross:
@HugoDF re PR validation - this would be a good idea, especially titles.

iamandrewluca:
@HugoDF 
> We might need to generate another token

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/authenticating-with-the-github_token#about-the-github_token-secret

HugoDF:
@iamandrewluca

> If you need a token that requires permissions that aren't available in the GITHUB_TOKEN, you can create a personal access token and set it as a secret in your repository

I believe the issue is with PR or comment access on forked repositories.

iamandrewluca:
That means that if you need more permissions that it offers `GITHUB_TOKEN`,  you should create your own.

HugoDF:
I agree, will sort out later today 👍.

In the meantime, the validation check is still working as expected, just no comments unless a maintainer opened the PR.

iamandrewluca:
Some references:
[...]

PR #1032, opened 2020-11-23
Revert "Revert "Use DangerJS for commenting""

HugoDF:
Reverts wesbos/awesome-uses#1027

I've reverted the use of Danger JS for commenting, in order for it to work, we'll need the `DANGER_GITHUB_API_TOKEN` @wesbos (see https://github.com/wesbos/awesome-uses/issues/478#issuecomment-626307220)

DangerJS integration doesn't work properly since we're missing a token I think see here: https://github.com/wesbos/awesome-uses/issues/999#issuecomment-714624048

When we've got the token in, we can merge this

### DIFF
--- a/.github/workflows/data-validate.yml
+++ b/.github/workflows/data-validate.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Install Dependencies
       run: npm install
 
-    - name: Validate data.js
-      run: node ./scripts/data-validate.js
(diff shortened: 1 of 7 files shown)


BlakeCampbells:
Hey, @HugoDF I'm going to close this out since it's been open for well over two years. 

If this is something, we want to add in the future feel free to reopen.

PR #1046, opened 2020-12-11
Add Stefan Zweifel

stefanzweifel:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -309,6 +309,31 @@ module.exports = [
     ],
   },
   {
+    name: 'Stefan Zweifel',
+    description:
+      'Full Stack Developer trying to make the web a better place. Working mostly with Laravel and Tailwind CSS.',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -308,6 +308,31 @@ module.exports = [
...
+  {
+    name: 'Stefan Zweifel',
+    description:
+      'Full Stack Developer trying to make the web a better place. Working mostly with Laravel and Tailwind CSS.',
+    url: 'https://stefanzweifel.io/uses/',
+    twitter: '@_stefanzweifel',
+    emoji: '✌️',
+    country: '🇨🇭',
+    computer: 'apple',
+    phone: 'apple',
```
HugoDF, line 320: I think the schema expects `iphone`

stefanzweifel: Whoops. Fixed that.

HugoDF (approved):

PR #1074, opened 2021-01-10
Add @thewannabeceo

marcdion:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -2,6 +2,24 @@
 // Data is validated against the following schema https://github.com/wesbos/awesome-uses/blob/master/scripts/utils.js#L53-L68
 module.exports = [
   {
+    name: 'Marc-Antoine Dion',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -1,6 +1,24 @@
 // keep it commonjs export
 // Data is validated against the following schema https://github.com/wesbos/awesome-uses/blob/master/scripts/utils.js#L53-L68
 module.exports = [
+  {
+    name: 'Marc-Antoine Dion',
+    description: 'Full Stack. Rookie cyclist. Part time blogger at @thewannabeceo.',
+    url: 'https://marcantoinedion.com/uses',
```
gatarelib, line 7: The site is empty, we would love to add it as soon as you add data to it.

gatarelib (changes requested):

gatarelib (approved):

PR #1089, opened 2021-01-28
Update Nico Bachner

nico-bachner:


src/data.js
```diff
@@ -6211,31 +6211,28 @@ module.exports = [
     name: 'Nico Bachner',
     description:
       'High School Student. Self taught Web Developer. Open Source enthusiast',
-    url: 'https://nicob.dev/#uses',
+    url: 'https://nicob.dev/uses',
```
gatarelib, line 6214: The page is 404

gatarelib (changes requested):

src/data.js
nico-bachner: Thanks. I'll fix that asap.

nico-bachner: Done.

gatarelib (approved):

PR #1090, opened 2021-01-28
fix(data): merge vscode tag (#637)

nico-bachner:


BlakeCampbells (approved):
Thanks for cleaning this up.

PR #1092, opened 2021-01-28
fix(data) merge express tag (#637)

nico-bachner:
tags merged:

- 'Express'
- 'ExpressJS'
- 'Express.js'


BlakeCampbells (approved):
I think this is a good change since `express.js` is how it's officially documented.

PR #1093, opened 2021-01-28
fix(data) merge go tag (#637)

nico-bachner:
Tags merged:

- 'Go'
- 'Golang'

into:

- 'Go'


BlakeCampbells (approved):
Screen Shot 2021-12-18 at 3 38 22 PM [Go Official Site](https://go.dev/doc/faq)

I think this change is good. Thanks for taking this on.

PR #1118, opened 2021-02-25
Add @onemohrtime

OneMohrTime:


src/data.js
```diff
@@ -11336,6 +11336,35 @@ module.exports = [
       'C#',
     ],
   },
+  {
+    name: 'Derek Mohr',
+    description: 'Front-end development by interest, full stack development by trade',
+    url: 'https://onemohrti.me/uses/',
```
gatarelib, line 11342: Error: Ping to "https://onemohrti.me/uses/" failed with status: 404

gatarelib (changes requested):

src/data.js
OneMohrTime: @gatarelib is there a way to run a re-check? It's definitely there, I wonder why it came back with a `404`...
http://onemohrti.me/uses

gatarelib (approved):

PR #1126, opened 2021-03-15
Add George Campbell

abersnaze:
George Campbell


src/data.js
```diff
@@ -6940,6 +6940,41 @@ module.exports = [
     phone: 'iphone',
     tags: ['Designer', 'Developer', 'Front End', 'JavaScript', 'React'],
   },
+  {
+    name: 'George Campbell',
+    description: 'Full stack engineer at Netflix',
+    url: 'https://gibr.net/uses',
```
gatarelib, line 6946: Ping to "https://gibr.net/uses" failed with status: 404

gatarelib (changes requested):

src/data.js
abersnaze: i fixed the redirect.

gatarelib (approved):

PR #1157, opened 2021-04-22
Add Dušan Simić

dusansimic:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -5888,6 +5888,17 @@ module.exports = [
     ],
   },
   {
+    name: 'Dušan Simić',
+    description: 'Computer Science student at University of Novi Sad, Serbia and open source software enthusiast.',
+    url: 'https://dusansimic.me/about/',
+    country: '🇷🇸',
+    twitter: '@dsnsmc',
+    emoji: '👨🏻‍💻',
+    computer: 'linux',
+    phone: 'android',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -5887,6 +5887,17 @@ module.exports = [
       'Musician',
     ],
   },
+  {
+    name: 'Dušan Simić',
+    description: 'Computer Science student at University of Novi Sad, Serbia and open source software enthusiast.',
+    url: 'https://dusansimic.me/about/',
```
gatarelib, line 5893: "url" with value "https://dusansimic.me/about/" fails to match the required pattern: /(use|uses|using|setup|environment|

gatarelib (changes requested):

src/data.js
dusansimic: Sorry to be a bother 😅 but is it really required for `uses` page to have one of those paths? I added things I use to my about page so that's why I've added that link.

src/data.js
```diff
@@ -5887,6 +5887,17 @@ module.exports = [
       'Musician',
     ],
   },
+  {
+    name: 'Dušan Simić',
+    description: 'Computer Science student at University of Novi Sad, Serbia and open source software enthusiast.',
+    url: 'https://dusansimic.me/uses',
```
gatarelib, line 5893: Great

gatarelib (approved):

PR #1178, opened 2021-06-04
add Michael Bonner

michaelbonner:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -2001,6 +2001,32 @@ module.exports = [
     ],
   },
   {
+    name: 'Michael Bonner',
+    description:
+      'Full stack JavaScript and PHP developer in Salt Lake City, USA',
+    url: 'https://michaelbonner/uses',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -2000,6 +2000,33 @@ module.exports = [
       'Movie Buff',
     ],
   },
+  {
+    name: 'Michael Bonner',
+    description:
+      'Full stack JavaScript and PHP developer in Salt Lake City, USA',
+    url: 'https://michaelbonner/uses',
```
gatarelib, line 2007: Error: Ping to "https://michaelbonner/uses" failed

gatarelib (changes requested):

src/data.js
michaelbonner: Wow... that was dumb :( Fixed now. Thanks!

HugoDF (approved):

PR #1227, opened 2021-08-23
add Bumhan Yu (@baadaa) && fix build failure on `Person.js`

baadaa:

### DIFF
--- a/src/components/Person.js
+++ b/src/components/Person.js
@@ -24,7 +24,7 @@ function useIntersectionObserver(ref) {
 
 export default function Person({ person, currentTag }) {
   const url = new URL(person.url);
-  const twitter = `https://unavatar.now.sh/twitter/${person.twitter.replace('@', '')}`;
+  const twitter = person.twitter ? `https://unavatar.now.sh/twitter/${person.twitter.replace('@', '')}` : null;
   const website = `https://unavatar.now.sh/${url.host}`;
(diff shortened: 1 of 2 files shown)


baadaa:
Just a quick recap of the two commits in this PR

1. Adding my info based on the existing `data.js` structure
2. Adding a ternary conditional for Twitter string to avoid breaking the build when there's no `twitter` value

src/components/Person.js
```diff
@@ -24,7 +24,7 @@ function useIntersectionObserver(ref) {
 
 export default function Person({ person, currentTag }) {
   const url = new URL(person.url);
-  const twitter = `https://unavatar.now.sh/twitter/${person.twitter.replace('@', '')}`;
```
baadaa, line 27: This string interpolation fails when it encounters a data entry with no `twitter` key in it—with the error `Cannot read property 'replace' of null`. My PR addresses it by a ternary operator to interpolate the string only when there's a `twitter` value.

baadaa:
Just checking in @HugoDF & @wesbos — many of currently open PRs fail during build at the same point (i.e. `Person.js` line 27), and this PR addresses the fix. Could you please review and merge? I believe it'll help resolve many of the open PRs.

image

HugoDF:
Thanks for fixing that

mreduar:
@HugoDF Do we need to send a new PR?

PR #1236, opened 2021-09-16
Add @Alfynodon

alfyn78:
Alfyn


BlakeCampbells:
@alfyn78 Will you please update your fork and see if that fixes this issue? 

You should see a screen like this on your fork, then you should just need to fetch upstream. I'm not sure if there's an easier way to do it on my end.
Screen Shot 2021-12-18 at 3 54 08 PM

BlakeCampbells (changes requested):
@alfyn78  Will you please update your fork and see if that fixes this issue?

You should see a screen like this on your fork, and then you should just fetch upstream. I'm not sure if there's an easier way to do it on my end.

Screen Shot 2021-12-18 at 3 54 08 PM

A change a while ago that fixed the issue in the build process.

BlakeCampbells (approved):
Thanks for merging in master!

PR #1249, opened 2021-10-11
feat: Makes tags navigable by keyboard

### LINKED ISSUE #527: Topics Selector is inaccessible
They are input radio buttons with labels, but hidden. Should be pretty simple to make accessible (tab-able).

g-traub:
One can navigate through the tags using "tab" and select one by pressing "enter"
Resolves #527

Guillaume


g-traub:
Thanks for reviewing it!

wesbos:
thanks a ton!

PR #1271, opened 2021-12-19
Fixing existing errors on master branch.

BlakeCampbells:
**Blake Campbell:** 

## Changes:
- [x] Going through and removing users with invalid URLs. 
- [x] Fixing other errors like country initials instead of emojis.


BlakeCampbells:
Going to remake this without the changes to package lock.

PR #1273, opened 2021-12-19
Removing "undefined" tag caused by missing phone types.

BlakeCampbells:
** Blake Campbell ** fixing undefined phone type caused by no-field set. 

@therealchiko will you please verify your phone type?
@seshaljain will you please verify your phone type as well?

Thanks!


therealchiko:
Thanks @BlakeCampbells! My phone type is correct here :)

PR #1278, opened 2022-01-02
Add Andrei Racasan

0xAMR:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -5640,6 +5640,27 @@ module.exports = [
     ],
   },
   {
+    name: 'Andrei Racasan',
+    description: 'Full Stack Developer with a passion for finding pragmatic solutions to technical challenges.',
+    url: 'https://www.andreiracasan.com/about',
+    emoji: '🌍',
+    country: '🇷🇴',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -5639,6 +5639,27 @@ module.exports = [
       'Python',
     ],
   },
+  {
+    name: 'Andrei Racasan',
+    description: 'Full Stack Developer with a passion for finding pragmatic solutions to technical challenges.',
+    url: 'https://www.andreiracasan.com/about',
```
BlakeCampbells, line 5645: Hey @0xAMR, will you please add a page with the requirements for the URL?

`https://www.andreiracasan.com/about`

`use|uses|using|setup|environment` at the end instead of about.

BlakeCampbells (changes requested):

src/data.js
0xAMR: My apologies, please see now.

BlakeCampbells (approved):
Wow! Record time! Thanks for fixing that so fast, getting it merged in now.

PR #1283, opened 2022-01-03
Added @DavidSmooke to uses.tech

davidsmooke:


src/data.js
```diff
@@ -70,6 +70,30 @@ module.exports = [
       'Blogger',
     ],
   },
+  {
+    name: 'David Smooke',
+    description: 'founder/ceo @hackernoon, leading the product team.',
+    url: 'https://www.davidsmooke.net/',
```
BlakeCampbells, line 76: Thanks, @davidsmooke. Will you update your site to have a uses page?
`url: 'https://www.davidsmooke.net/'`, needs to follow the format of

`use|uses|using|setup|environment` at the end.

After you fix that then we can get this merged in!

BlakeCampbells (changes requested):

src/data.js
davidsmooke: Hey @BlakeCampbells! just looked through a couple examples and made: https://www.davidsmooke.net/uses does that work?

BlakeCampbells: That works @davidsmooke!

BlakeCampbells (approved):

PR #1294, opened 2022-01-24
UPDATE Patrick Obermeier

Pobermeier:
Hello!

I would like to update my domain as it has recently changed.

Thanks in advance!

Best,
Patrick


BlakeCampbells (approved):
Awesome! Thanks for updating it @Pobermeier

PR #1295, opened 2022-01-26
Adding Saurav Khare

SauravKhare:
Adding Saurav Khare


BlakeCampbells (changes requested):
Will you please add a `uses` URL to your website? It should explains what software/hardware you use.

src/data.js
```diff
@@ -1,6 +1,34 @@
 // keep it commonjs export
 // Data is validated against the following schema https://github.com/wesbos/awesome-uses/blob/master/scripts/utils.js#L53-L68
 module.exports = [
+  {
+    name: 'Saurav Khare',
+    description: 'Software Engineer & Frontend Developer',
+    url: 'https://sauravkhare.xyz/uses',
```
BlakeCampbells, line 7: It looks like this is an empty page right now. Will, you see if you can fix it @SauravKhare?

BlakeCampbells (changes requested):

src/data.js
SauravKhare: The issue is fixed.

BlakeCampbells (approved):

BlakeCampbells:
@SauravKhare Awesome thanks for fixing that!

PR #1308, opened 2022-02-19
add Simon Lee

simonhlee97:


src/data.js
```diff
@@ -13233,6 +13233,34 @@ module.exports = [
...
+      'VueJS',
+      'SCSS',
+      'TypeScript',
+      'JamStack',
+      'Node.js',
+      'MongoDB',
+      'Postgres',
+      'UI Developer',
+      'web development',
+      'golf handicap: 15',
```
BlakeCampbells, line 13260: Hey, @simonhlee97 will you remove some of these unique tags and add them to your bio if you'd like them to show?

``` JavaScript
  const tags = Object.entries(counts)
    .sort(([, countA], [, countB]) => countB - countA)
    // Only show the tag if this topic has 3 or more people in it
    .filter(([, count]) => count >= 3)
    .map(([name, count]) => ({ name, count }));
```

Right now only tags with 3+ people show.

BlakeCampbells (changes requested):

src/data.js
simonhlee97: removed those tags. Thank you.

BlakeCampbells (approved):

PR #1309, opened 2022-02-20
Cleanup Feb 2022

BlakeCampbells:
Blake Campbell: Cleaning the following sites either no longer exist or the `/uses` page is gone. 

If you would like to be re-added, please submit a new PR. I've tried to separate each removal into individual commits to keep it accessible to see what was removed. 

@famiclone 
@kushanshamika 
@boardfish 
@Mozart409 
@samirmhsnv 
@joshfarrant
@jakeherp 
@OneMohrTime

PR #1312, opened 2022-02-20
Re-adds Jacob Herper

jakeherp:
My Domain has changed recently and it was removed as part of #1309 
Re-adding with updated URL.


BlakeCampbells (approved):
Thanks for updating @jakeherp! I love the purple and black for your site!

PR #1321, opened 2022-03-09
Update EosisWasTaken

EosisWasTaken:
I got a website now, so I am updating this since my old uses page is not relevant.


EosisWasTaken:
Interesting, the check says it failed to ping my site but it works on my end... Is that because the site is brand new (I bought the domain name a couple of hours ago, and the hosting is done by GH Pages) ?

BlakeCampbells (changes requested):
I can't seem to hit your new URL either. 

We can wait a little bit and see if it's still updating, it could be because the domain is brand new. It should have updated by now though.

EosisWasTaken:
Alright it works now, I had to modify something in the DNS config on the domain name. Now everything is ready, we can run the tests again.

BlakeCampbells (approved):

BlakeCampbells (approved):
Looks like it's passing now, thanks for fixing that!

PR #1332, opened 2022-03-20
Update Emma Goto

emgoto:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -5827,12 +5827,11 @@ module.exports = [
   },
   {
     name: 'Emma Goto',
-    description:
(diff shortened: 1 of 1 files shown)


BlakeCampbells (changes requested):
Hey @emgoto will you please update your PR to only modify your section?

It probably automatically from ESLint or Prettier

emgoto:
My bad! I assumed that was the pre-commit hook for the repo so I left it in. Have removed now. Thanks!

BlakeCampbells (approved):
Perfect, thanks @emgoto!

PR #1338, opened 2022-04-07
Add @nikolalsvk

nikolalsvk:


src/data.js
```diff
@@ -2941,6 +2941,35 @@ module.exports = [
...
+      'JavaScript',
+      'CSS',
+      'React',
+      'Ruby',
+      'Ruby on Rails',
+      'Developer',
+      'Speaker',
+      'Blogger',
+      'Teacher',
+      'Motorcycle',
```
BlakeCampbells, line 2970: Looks great, just so you know Tags won't appear unless multiple people have them. "Motorcycle" might not show up.

BlakeCampbells (approved):

src/data.js
nikolalsvk: Thanks 🙇 I saw it from the person above me so I put it for fun 😂

PR #1343, opened 2022-04-16
Introduce JSDocs for data.js

BenDMyers:
This pull request doesn't actually introduce any new uses pages. Instead, it adds some JSDoc comments to `data.js`. These comments will provide autocomplete hints if a contributor is adding their entry using an editor such as VS Code. It won't perform any sort of validation — and mentions of the real validator are maintained — but it might help people more easily fill out enum fields like `computer` or `phone` by presenting them with their options as they type.


BlakeCampbells (commented):
@wesbos would this change conflict with any of the updates you want to make?

BlakeCampbells (approved):

PR #1348, opened 2022-05-01
Adding Lazar Nikolov

nikolovlazar:


BlakeCampbells (approved):
Looks great! I'll have to check out Warp as an iterm replacement.

PR #1364, opened 2022-05-22
Add Stuart McColl

stuartmccoll:
Stuart McColl


BlakeCampbells (approved):
I like the minimalism, you could also add an `emoji: '👟',` to your section if you'd like.

It doesn't seem like it's required by the current testing though.

PR #1367, opened 2022-05-24
chore: update Nico Bachner

nico-bachner:
- updated url
- updated description
- added twitter handle
- removed emoji
- updated tags

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -7689,31 +7689,41 @@ module.exports = [
     tags: ['Developer', 'Designer', 'Front End', 'JavaScript', 'React', 'Bots'],
   },
   {
+    // required tags
     name: 'Nico Bachner',
-    description:
-      'High School Student. Self taught Web Developer. Open Source enthusiast',
-    url: 'https://nicob.dev/uses',
-    emoji: '🎿',
+    description: 'Student. Developer. Entrepreneur.',
+    url: 'https://nicobachner.com/uses',
     country: '🇱🇺',
+    
+    // optional tags
+    twitter: '@nico_bachner',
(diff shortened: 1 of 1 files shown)


BlakeCampbells (changes requested):
Hey @nico-bachner will you please remove comments? Just to keep each person uniform in the file.

nico-bachner:
May I ask why exactly the comments are an issue? 
I'm not sure I understand the reasoning behind your proposed change, given that there are already large variances in the order and formatting of the object properties.

From my perspective, I see two reasons why the comments are a good addition:

1. The comments clarify for me how the object properties map to the person information schema[^1], something that currently isn't obvious. Initially, finding out this information required me to go through the source code, and if I were to have to go through that process again every time I make an update to my information, it would be a bit of a waste of time.
2. The comments might be useful for new people submitting their first PR. If they copy my person information object and modify the content, the way I did when submitting my first PR here, they will have a better understanding of which properties are required or optional.

Therefore, unless adding comments will in some way break the build, I would suggest keeping the comments.

[^1]: https://github.com/wesbos/awesome-uses/blob/b959f850fe10543b23a72ac49fdecde9fd825e5e/scripts/utils.js#L53-L68

BlakeCampbells:
Good points! 
I went ahead and merged [PR 1343](https://github.com/wesbos/awesome-uses/pull/1343/files).
If you'd like to expand on that to specify what's `required` vs `optional` I think the top of `data.js` file makes sense. 

I'll go ahead and merge this.

BlakeCampbells (approved):

PR #1371, opened 2022-06-01
Adding Nick Reutlinger

NickRTR:
Add Nick Reutlinger


BlakeCampbells (approved):
Made a small update to phone section, but looks good!

PR #1383, opened 2022-06-14
Update Marc Littlemore details

MarcL:
## Summary
- Update Marc Littlemore role details - I now work at Netlify
- Added a trailing slash to the `/uses/` URL to avoid a redirect 😄 

Thanks!


BlakeCampbells (approved):
Congrats on the move to Netlify! I've heard nothing, but great things from developers there!

PR #1417, opened 2022-08-15
Add myself back in, with updated URL

jmathias:


BlakeCampbells (approved):
Thanks for updating @jmathias! The Sennheiser 599s are amazing!

PR #1419, opened 2022-08-23
Updates @fourjuaneight's website

fourjuaneight:
Auto-renew was off and I lost the domain 😅 I heard that all the cool kids are using those `xyz` TLDs.


BlakeCampbells (approved):
@fourjuaneight Thanks for being on top of it and updating it.

PR #1428, opened 2022-09-10
Add @weakish

weakish:


BlakeCampbells (approved):
Thanks @weakish! I think you might be the first person with a `Wear` section, it's a great addition!

PR #1439, opened 2022-10-02
📦 NEW: add link check action

TechWiz-3:
In this PR, I have added a workflow which checks links and fails if any of them are broken.

Disclaimer: this action is my own, the repo is [here](https://github.com/TechWiz-3/linksafe/tree/fast).

Because of the large volume of links, I'm using the 'fast' version/branch which simply requires a GitHub token (with no extra perms) to be stored as a repo secret under the name  `TOKEN`. This is for requests to the GitHub API to avoid rate-limits. More setup info [here](https://github.com/TechWiz-3/linksafe/tree/fast#setup)

You can view the logs of the workflow on my fork [here](https://github.com/TechWiz-3/awesome-uses/actions/runs/3165357794/jobs/5154323420). **51** links were found to be broken.

Let me know if you'd like me to change anything, any feedback is appreciated either way :)

Screen Shot 2022-10-02 at 3 28 09 pm

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -1,0 +1,15 @@
+name: Link-check
+on:
+  schedule:
+    - cron: '0 0 * * 1'  # every monday
+  push:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run linksafe
+        uses: TechWiz-3/linksafe@fast
+        env:
+          TOKEN: ${{ secrets.TOKEN }}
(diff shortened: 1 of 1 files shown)


BlakeCampbells (commented):
@TechWiz-3 we currently validate urls when the PR is made, but only for additions. That way we don't hold up PRs for new contributions based off old urls expiring. 

Inside of `scripts/` there's one to go test every url. I try to run that every month or two to clean up expired links.

I appreciate the thought, but this could get frustrating for first timers.

TechWiz-3:
I accept your decision and thanks for the response but fyi `on: push` only triggers for commits to the repo, not pull requests.

PR #1495, opened 2023-01-07
Add Nikita Karamov

kytta:


BlakeCampbells (approved):
Thanks for contributing! Zsh is such a huge terminal upgrade

PR #1496, opened 2023-01-07
Update unavatar URL

kytta:
So, I've noticed that none of the profile pictures work. I have visited unavatar.now.sh, but it's inactive at the moment. It seems like they have a new URL: unavatar.io. This PR replaces the old URL with the new one.


kytta:
I see it was fixed in d15c128a11c5447bf53250e766a70f3cd24855b6 & 74af76a9cd86aa68d4adf85e805eafc6e2583a3c, closing

PR #1526, opened 2023-02-19
Add Adil Haddaoui

adilhaddaoui:
*Adil Haddaoui*


BlakeCampbells (approved):
Thanks for contributing @adilhaddaoui, big fan of ohmyzsh!

PR #1538, opened 2023-02-22
Fix BackToTop link not showing after scrolling past 25% of the page

theMosaad:


wesbos:
thanks! Totally forgot I broke this when removing the styled components

PR #1543, opened 2023-02-23
Testing content-visibility: auto

wesbos:
Apparently this should defer the rendering until they are scrolled into view?


wesbos:
wow that significantly improved it  - still too slow, but nice trick!

PR #1559, opened 2023-03-21
Reduce data usage by sorting and filtering on the frontend

theMosaad:
### Problem

Data for people is being refetched every time a user hovers over or click on a tag. That's usually after all data has been loaded when the user lands on the homepage for the first time.

### Solution

Sort and filter the fetched people on the frontend after fetching the data one.

This reduces data usage and provides faster page transitions.


jovylle:
Hi @theMosaad there seems to be new commits conflicted here.

BlakeCampbells:
Closing, but this is a great idea for speeding up the site along with #1740

Issue #1558, opened 2023-03-19
Image loading CDN 404s


Hacksore:
Hi Folks! I wanted to start this discussion to talk about the 3 dimensions about how images are served 😅. 

I saw many failures on the CDN today so figured it might be worth discussing options to solve this.

```
Dimension 1: images.weserv.nl (which this is more of a CDN for both the image and fallback)
Dimension 2: unavatar.io/Hacksore
Dimension 3: unavatar.io/boult.me&w=100&l=9&af&il&n=-1
```

Errors seeen
```
GET https://images.weserv.nl/?url=https://unavatar.io/MRebehn?fallback=https://unavatar.io/mattaz.com&w=100&l=9&af&il&n=-1 404
GET https://images.weserv.nl/?url=https://unavatar.io/kmorope?fallback=https://unavatar.io/camiloromero.dev&w=100&l=9&af&il&n=-1 404
GET https://images.weserv.nl/?url=https://unavatar.io/trovster?fallback=https://unavatar.io/www.trovster.com&w=100&l=9&af&il&n=-1 404
GET https://images.weserv.nl/?url=https://unavatar.io/BenDMyers?fallback=https://unavatar.io/benmyers.dev&w=100&l=9&af&il&n=-1 404
```

wesbos:
hmmm yeah

The URL stack is actually:

Dimension 1: images.weserv.nl?url=XXX&w=100&l=9&af&il&n=-1
Dimension 2: unavatar.io/Hacksore
Dimension 3: unavatar.io/boult.me

Unavatar made the twitter images paid, so that is why the quality has gone down.

I think the issue is that the weserv.nl url isn't falling back

Broken:

https://images.weserv.nl/?url=https://unavatar.io/MRebehn?fallback=https://unavatar.io/mattaz.com

Working:

https://unavatar.io/MRebehn?fallback=https://unavatar.io/mattaz.com


Perhaps it's a URL encoding issue?

Hacksore:
Can we just hotlink Github avatars 😬?

wesbos:
Half tempted to write a scraper that would render twitter and get the profile image..

Hacksore:
I'd say a long term solution would be to require the Github user id with the data object?

Then we can hotlink their image:
`https://avatars.githubusercontent.com/u/<userId>?s=80`

However, this makes it hard to go and fix all existing users. Maybe it could just be a manual one time migration effort (automation might be not worth it in this case)?

[...]

PR #1564, opened 2023-03-25
Add fallback image

Hacksore:
This is just a simple stop gap to add a fallback image the persons avatar until a long term solution is done for #1558.

@wesbos Let me know if you have any thoughts?


wesbos:
looks good -thanks a ton

PR #1578, opened 2023-04-26
Update uses page URL for Vincent Lejtzén

lejtzen:
The URL for my uses page has changed — I'm just updating it here to match the new one.


BlakeCampbells (approved):
Thanks for the update!

PR #1601, opened 2023-06-15
adding information

herbras:


BlakeCampbells (approved):
Love that you included font family. Thanks for contributing! @herbras

PR #1602, opened 2023-06-16
Add Jovylle Bermudez

jovyllebermudez:
Adding Jovylle Bermudez A listener of hundreds of syntax episodes


BlakeCampbells (dismissed):
It looks like your site is using white text on a white background. Will you adjust it and then we can get it merged in. 
Screen Shot 2023-06-21 at 11 08 09 PM

gatarelib:
> 

I can see that it was rectified.

gatarelib (approved):

jovyllebermudez:
> It looks like your site is using white text on a white background. Will you adjust it and then we can get it merged in. Screen Shot 2023-06-21 at 11 08 09 PM

Apologies for the late reply, so Weird I missed the notification for this request, not on my notification list at all.

Thanks for this heads up, the Dark mode CSS classes never expected to activate. I'll continue setting up its dark mode or remove it so won't activate.

jovyllebermudez:
> I can see that it was rectified.

Thank you for merging. I found what's causing it to use a white color font.

PR #1614, opened 2023-07-06
Add Kamuri Amorim

kamuridesu:


BlakeCampbells (approved):

BlakeCampbells (changes requested):
Will you please update your uses to a page or repo on github like:
Example: 
https://github.com/preetsahil/uses

PR #1627, opened 2023-07-11
Remove duplicated tags

jcubic:
This is against the rule but I wanted to fix some duplicated tags with different spellings.
I hope it's ok.


BlakeCampbells (approved):
Awesome thanks for the tag clean up!

PR #1637, opened 2023-07-30
Add Tim Bachmann

Tiim:


BlakeCampbells (approved):
Thanks for contributing @Tiim! I keep seeing great things about Svelte, I'll have to test it out at some point.

Tiim:
It's definitely worth checking out. 😀

PR #1638, opened 2023-08-02
Add Sho Koike

putcut:


BlakeCampbells (approved):
Thanks for contributing @putcut!

PR #1641, opened 2023-08-08
Add michaelmior

michaelmior:


BlakeCampbells (approved):
I've seen so many people use Obsidian. I'll have to give it a go eventually

michaelmior:
I've only started using it within the past year, but it's been great! I feel like I'm just scratching the surface when it comes to available plugins.

PR #1655, opened 2023-08-26
Updated URL for eklausmeier.goip.de

eklausme:
Just changed URL for eklausmeier.goip.de, no new entry.


BlakeCampbells (approved):
Thanks for updating!

PR #1663, opened 2023-09-16
Adding Theodoros Ploumis

theodorosploumis:
https://www.theodorosploumis.com/en/uses

Thank you.

PR #1666, opened 2023-09-22
Update utils.js

gatarelib:
Allowing both null and 'none' for what kind of phone one has for more flexibility and compatibility in the schema.
A case for PR [#1663](https://github.com/wesbos/awesome-uses/pull/1663)


wesbos:
ah, I don't think people don't actually have a phone. gonna reject this one

PR #1680, opened 2023-10-13
Gurdeep Singh - Happy Hacking

Gurdeep99:


jovyllebermudez:
why you naming it like that super sus

BlakeCampbells:
> why you naming it like that super sus

It was probably for Hacktober Fest.

src/data.js
```diff
@@ -21,6 +21,64 @@
  * Keep this a commonjs export.
  */
 module.exports = [
+  {
+    name: 'Gurdeep Singh',
+    description: 'Full Stact Developer in PHP and ReactJS (MERN)',
+    url: 'https://www.gurdeep.net/uses',
```
BlakeCampbells, line 27: Two changes:
- Make sure your uses page is live
- Make sure the country is the flag emoji instead of IN

BlakeCampbells (changes requested):

PR #1691, opened 2023-11-09
Add Guru Das Srinagesh

guru-das-s:
Add myself to the list.


BlakeCampbells (approved):
Thanks for contributing @guru-das-s, Seiko watches are some of the best!

PR #1693, opened 2023-11-22
Add Michael Burkhardt

mihobu:


BlakeCampbells (approved):
Thanks for contributing @mihobu! Love that you included a kettle on your site, completely agree and end up using it almost daily!

PR #1715, opened 2024-01-09
Add Hector Aguilar

hectoraldairah:

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -17972,6 +17972,26 @@ module.exports = [
     ],
   },
   {
+    name: 'Hector Aguilar',
+    description:
+      'A Software Engineer that loves Web development and CLI sofware',
+    url: 'https://bitbyte.blog/uses',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -17971,6 +17971,26 @@ module.exports = [
       'Problem Solver',
     ],
   },
+  {
+    name: 'Hector Aguilar',
+    description:
+      'A Software Engineer that loves Web development and CLI sofware',
+    url: 'https://bitbyte.blog/uses',
+    emoji: '🐤',
+    computer: 'macos',
```
gatarelib, line 17980: "computer" must be one of [apple, windows, linux, bsd]

gatarelib (changes requested):

src/data.js
hectoraldairah: Thank you for the correction!

gatarelib (approved):

PR #1746, opened 2024-02-12
Add Hugh He

plushugh:

**Add your name to verify:** Hugh He

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -14116,6 +14116,22 @@ module.exports = [
     ],
   },
   {
+    name: 'Hugh He',
+    description: 'FOSS Developer, Student',
+    url: 'https://github.com/plushugh/plushugh/blob/main/USES.md',
+    twitter: '@plushugh',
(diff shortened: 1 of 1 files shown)


src/data.js
```diff
@@ -14115,6 +14115,22 @@ module.exports = [
       'Node.js',
     ],
   },
+  {
+    name: 'Hugh He',
+    description: 'FOSS Developer, Student',
+    url: 'https://github.com/plushugh/plushugh/blob/main/USES.md',
```
gatarelib, line 14121: Hey! The 'url' pattern requires 'use', 'uses', 'using', 'setup', or starting with 'https://gist.github.com/'. Consider updating the URL.

gatarelib (changes requested):

src/data.js
plushugh: Thanks, will add a correct link in a moment

Issue #1593, opened 2023-06-07
feature request: swap to, or allow mastodon links


miclgael:
Less Musk, more Tusk.

wesbos:
Would be nice to make it a bit more flexible - will def think on how to do this!

miclgael:
Maybe if the string you set for social has two @ symbols it could be treated as fediverse?

PR #1757, opened 2024-02-23
Add jordan.hofker.org/uses/

jhofker:
Not sure if a Mastodon handle will work, but let's find out!

✅ You've read and understood the above
**Add your name to verify:** Jordan Hofker


gatarelib (changes requested):
The Twitter handle provided, "@jhofker@hachyderm.io", doesn't fit the expected format. It should only contain letters, numbers, or underscores, and be between 1 to 15 characters long.

jhofker:
Updated to just use Twitter handle. Would be great to get #1593 going to support Mastodon!

jhofker:
Hey @gatarelib when you have a moment, can you re-review? Thanks!

BlakeCampbells (approved):
Thanks for contributing!

PR #1774, opened 2024-03-13
Add Kevin Gimbel

KevinGimbel:
**Add your name to verify:** Kevin Gimbel


src/data.js
```diff
@@ -8283,6 +8283,25 @@ module.exports = [
       'PHP',
       'SQL',
     ],
+  },{
+    name: 'Kevin Gimbel',
+    description: 'DevOps Engineer who loves Rust + web dev',
+    url: 'https://kevingimbel.de/uses',
+    twitter: '',
```
gatarelib, line 8290: "twitter" is not allowed to be empty

gatarelib (dismissed):

src/data.js
KevinGimbel: Oh, so I can't join if I don't have a twitter account anymore?

BlakeCampbells: Sorry for the delay, `twitter: '',` is invalid, but if you remove the line it is valid.

KevinGimbel: I see, I removed it thanks! :)

PR #1779, opened 2024-03-20
Add R74n

slweeb:
Updated data.js to include R74n, a personal project website.


BlakeCampbells (approved):
Thanks for contributing! @slweeb

PR #1784, opened 2024-04-13
Update data.js

mledan:


src/data.js
```diff
@@ -18206,6 +18206,40 @@ module.exports = [
       'Bootstrap',
       'Bulma',
     ],
+  },  
+  {
+    name: 'Mladen Milesic',
+    description:
+      'Passionate Electrical Engineer gone Software. I relish the art of crafting and tinkering with both digital and physical creations.',
+    url: 'https://www.mladenmilesic.com/uses/',
```
BlakeCampbells, line 18214: Will you re-request after you populate the `uses` page?

BlakeCampbells (changes requested):

src/data.js
mledan: Hey @BlakeCampbells, I've populated my uses page, will be uploading some tasteful images to go alongside those as well! I see this is still open, any other actions needed to complete it?

BlakeCampbells:
Thanks for contributing @mledan!

PR #1797, opened 2024-05-22
country flag character code off center fix

jovyllebermudez:
Quick fix for not center flag codes due to the character themselves not being centered


jovyllebermudez:
Hi Wes, @wesbos, we have no longer need my code here, the country flags are now actual flags.

PR #1806, opened 2024-06-02
Adding Swastik Baranwal

Delta456:
Adding **Swastik Baranwal**


Delta456:
image


How do I display my profile picture? I thought it would automatically take from my GitHub profile.

wesbos:
hmm yah that seems to have been removed at some point. Let me fix that

wesbos:
Oh - we don't collect github profile data. That is why. TWitter is best, but the unavatar rate limits to 50 requests per day, and that isn't enough for us

Delta456:
> Oh - we don't collect github profile data. That is why. TWitter is best, but the unavatar rate limits to 50 requests per day, and that isn't enough for us

Afaik Twitter API is paid so it may be better to default to GitHub if that is something which can be considered.

PR #1808, opened 2024-06-05
Adding Jorge Cortés

jorgecortesdev:
Jorge Cortés


BlakeCampbells (approved):
Thanks for contributing @jorgecortesdev, I like the hover animation for the cards on your site!

PR #1823, opened 2024-07-02
Adding S. M. Abdul Aziz

sayemon10:
https://www.sayemon10.com/uses


BlakeCampbells (approved):
Thanks for contributing! 
```
Attention is the only true currency we have in life.
Thanks for gifting me yours.
```

PR #1825, opened 2024-07-09
Add Rachel Cantor

rachelslurs:
👋


BlakeCampbells (approved):
Thanks for contributing @rachelslurs! I haven't used Cherry MX Blue switches yet, but I love the red ones.

PR #1828, opened 2024-07-16
update Mahesh Rijal

maheshrijal:
Edited myself. 
First added in https://github.com/wesbos/awesome-uses/pull/1802


BlakeCampbells:
Thanks for the update!

PR #1832, opened 2024-07-30
Add Chris Funderburg

bocan:

**Add your name to verify:** Chris Funderburg


BlakeCampbells:
Thanks for contributing! Having the Everyday Carry section was a great addition.

PR #1878, opened 2024-11-15
Add support for bluesky urls with one test case

whitep4nth3r:
Hello! 

I've added support for Bluesky URLs in profile data.

To avoid complications, I've made it so that if you don't have Twitter,  you don't have Mastodon, but you do have Bluesky, we show that link (which is in line with how the logic was written previously in comparing Twitter and Mastodon).

image


wesbos:
sweet thanks- we should probably add the ability to have multiple socials, but this is great for now.

PR #1885, opened 2024-11-28
Add Charles Harwood

charles4221:
**Add your name to verify:** Charles Harwood


BlakeCampbells:
Thanks for contributing @charles4221!

PR #1897, opened 2025-01-03
Add @obifortunebleh

ickynavigator:
**Add your name to verify:** Fortune Obi


src/data.js
```diff
@@ -16477,6 +16477,40 @@ module.exports = [
       'JAMstack',
     ],
   },
+  {
+    name: 'Fortune Obi',
+    description:
+      'Web developer living in Canada, strangely obsessed with Typescript and React.',
+    url: 'https://www.obifortune.com/uses',
```
BlakeCampbells, line 16484: Looks like this returns a 404 page right now

ickynavigator: Sorry about that! I was migrating my website and hadn't moved that page it yet.

It should be up now!

BlakeCampbells:
Thanks for the quick fix!

PR #1907, opened 2025-01-14
Add Jack Jackson

scubbo:
## Final Requirements

**Add your name to verify:** Jack Jackson


BlakeCampbells:
Thanks for contributing @scubbo

PR #1915, opened 2025-02-02
Add Abyan Raditya

byanrkh:


src/data.js
```diff
@@ -13628,6 +13628,26 @@ module.exports = [
       'Bootstrap',
     ],
   },
+  {
+    name: 'Abyan Raditya',
+    description: 'Hey! I am a Frontend web developer.',
+    url: 'https://www.byanrkh.my.id/blog/uses',
```
BlakeCampbells, line 13634: Looks like this redirects to 404 currently

PR #1934, opened 2025-03-03
Adding Jared Rigby

jazibobs:


BlakeCampbells:
Thanks for contributing @jazibobs!

PR #1938, opened 2025-03-13
Add Karel De Smet

carlos-ds:


carlos-ds:
Hi @BlakeCampbells or @wesbos: there's a check failing due to still using `actions/cache@v1` in [data-validate.yml](https://github.com/wesbos/awesome-uses/blob/5c7428c075428f63db37795c8544d2e4445b45e1/.github/workflows/data-validate.yml#L20) and/or [populate-readme.yml](https://github.com/wesbos/awesome-uses/blob/5c7428c075428f63db37795c8544d2e4445b45e1/.github/workflows/populate-readme.yml#L23). 

This deprecation / breaking change was mentioned in a [recent notice](https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down). Any chance you could fix this so we can get this PR merged? According to the [documentation](https://github.com/actions/cache/discussions/1510), updating to the latest version should be sufficient and you should not expect any breaking changes. If you want, I could open a PR myself but I assume you might somehow want to test this first?

BlakeCampbells:
It took a little time, I ended up getting us to Actions 4.x, luckily not a huge lift. I just needed to set aside some time and now I'm trying to clear the backlog a little at a time.

Thanks for contributing!

PR #1961, opened 2025-05-15
Add @jeffpaul

jeffpaul:
Adds reference to https://jeffpaul.com/uses/ and related metadata.


BlakeCampbells:
@jeffpaul thanks for contributing!

PR #1872, opened 2024-10-26
Add bluesky socials support

kendallm:
Kendall Morgan

## Description

Added support for bluesky social links. When a person defines multiple socials they are resolved in the priority of twitter -> bluesky -> mastodon.

The change was tested by adding my personal uses page and ensuring that bluesky resolves before my mastodon page for my social link.

**Note**: This change also fixes an issue with `unavatar.io` profile image generation. It seems like they changed their url structure and now require `unavatar.io/x/{username}` to render X/Twitter profile images. 

Resolves #1723

PR #2003, opened 2025-09-25
Fix duplicate Bluesky links 

csmith:
If a person has only a Bluesky link, it was displayed twice. It looks like support was added in two independent PRs: #1872, and #1878.

You can see an example of the issue on the single profile displayed here: https://uses.tech/like/Simplicity

I've also tweaked how the handles are normalised. Previously one block removed `@`s in the URL, then blindly chomped off the first character in the display name, while the second assumed there was no `@` prefix. There are a few entries that have an `@`, so I've stuck with just removing it in both places, as that'll work in all current cases.

### DIFF
--- a/src/components/Person.js
+++ b/src/components/Person.js
@@ -90,20 +90,6 @@ export default function Person({ person }) {
           </div>
         )}
 
-        {/* If they have a bluesky, and no twitter/mastodon, show that */}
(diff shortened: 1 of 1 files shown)


ar1ja (approved):
This indeed does fix the issue.

BlakeCampbells:
Thanks for the fix!

PR #2080, opened 2026-03-05
Add Tony Dehnke

tonydehnke:
Adding my /uses page — digital nomad & non-technical founder running a business from a backpack with Apple Vision Pro as a virtual display.

URL: https://tonydehnke.com/uses


tonydehnke:
@wesbos - it looks like a few of us from around this PR date range had failing CI - could be do to URL format of the test?   I've updated this so it should pass now, but the issue might have affected a few others..  possibly  trailing / in url?

tonydehnke:
@wesbos - any idea why this isn't passing and merging?

BlakeCampbells:
@tonydehnke, I think it's an issue with robots.txt or something similar. I'll dig into it when I have more free time. Thanks for your patience.

PR #2103, opened 2026-03-29
Add David M. Coleman

nookwerks:
I hope I'm doing this right.
This is I think only my second pull request ever, so please forgive me if I did it wrong.


nookwerks:
Um, about that. I gave up thinking it would get added. I ended up rebranding myself as Matt Coleman with the domain mattology.us. (Matthew is my middle name; never really liked David).
And yes, I know my github still says David and Nookwerks, but I've moved over gitlab so I never bothered to change it.

PR #2104, opened 2026-04-01
Update Nelson Lai

nelsonlaidev:
- Update my name
- Update to use the new domain
- Update twitter handle
- Update phone type


BlakeCampbells:
Thanks for the update!

PR #2118, opened 2026-04-20
Add Thxie

ThxieX:
Added Thxie to the /uses list. Website: https://thxie.com/about#uses

## Final Requirements

**Add your name to verify:** Thxie


BlakeCampbells:
Thanks for contributing @ThxieX!

PR #2137, opened 2026-05-22
Add Genís Matutes

ItsGenis:


BlakeCampbells:
Logitech MX Master is the best mouse ever! Thanks for contributing!

PR #2140, opened 2026-05-25
Add Prince Sinha

prince-vishal:
Adds Prince Sinha's /uses page: https://theprincevishal.in/uses


BlakeCampbells:
Thanks for contributing @prince-vishal!

PR #2143, opened 2026-05-30
Add Andri Óskarsson

andrioid:
Added Andri Óskarsson. Website at [andri.dk](https://andri.dk) and uses page at https://andri.dk/uses


BlakeCampbells:
Please make sure the commit is clean and not formatting the entire file @andrioid. Thanks!

PR #2147, opened 2026-06-02
Add Curtis Timson

curtiscde:
Adds Curtis Timson to the list.

- **Uses page**: https://www.curtiscode.dev/uses


BlakeCampbells:
Thanks for contributing @curtiscde!

PR #2160, opened 2026-06-28
Add Todor Aleksandrov

todor147:
Added new profile for Todor Aleksandrov with details and tags.


BlakeCampbells:
Thanks for contributing @todor147 

I think you might be the first Sommelier!

PR #2166, opened 2026-07-08
Add Angad Singh to data.js

singhangadin:
Added a new entry for Angad Singh with details on his expertise and social links.


BlakeCampbells:
Thanks for contributing @singhangadin!

PR #2169, opened 2026-07-12
Add Andrew Minga

amingacspire:
Adding my /uses page: https://andrewminga.com/uses

- URL ends in `uses` per the guide
- Entry inserted mid-array (not at the end) per the guide
- Tags reuse existing tag spellings


BlakeCampbells:
Thanks for contributing @amingacspire!

PR #2174, opened 2026-07-31
Add Felix Beßler

felixbessler:
Adding my /uses page: https://www.felix-bessler.de/uses

- Full-Stack Developer from Germany
- Added a single entry to `src/data.js` (inserted mid-array, not at the end)
- README not touched (auto-generated)


BlakeCampbells:
Thanks for contributing @felixbessler, I'm a big fan of Vuejs too

felixbessler:
> Thanks for contributing @felixbessler, I'm a big fan of Vuejs too

Thanks for letting us show our work. 
Sadly, I mostly have to work in Vuejs 2 but it is way better than jQuery, what I was doing before. 😅

PR #2179, opened 2026-08-13
Add Melih Arık

meliharik:
Adding my /uses page: https://meliharik.dev/uses

- [x] Added myself somewhere random in the array, not at the end
- [x] `src/data.js` still parses and validates
- [x] The page is live and reachable


BlakeCampbells:
Thanks for contributing!

PR #2200, opened 2026-09-15
Add Hoang Nguyen

folliehiyuki:
Adds my personal `/uses` page to the list.

URL: https://www.folliehiyuki.com/misc/uses.


BlakeCampbells:
@folliehiyuki thanks for contributing! Your devices having names from Lord of The Rings is great!

